PR #4, opened 2020-04-24
Support reading files from local disk

### LINKED ISSUE #1: Support accessing notebooks on the local disk in parallel to notebooks stored in GCS
The initial implementation of the GCS ContentsManager only supports reading and writing notebooks to GCS.

This means that if a user configures their Jupyter server to use the GCS ContentsManager, then they can no longer access notebooks stored on their local disk.

This issue proposes to change that by introducing a virtual top-level directory that lets the user choose between GCS and their local disk.

For example, if a user has a GCS bucket named `<bucket>`, then their directory layout might look like:

```
/
|
+- Local Disk/
|            |
|            +- Notebook1.ipynb
|            |
|            +- Notebook2.ipynb
|
+- gs://<bucket>/
                |
                +- Notebook3.ipynb
```

ojarjur:
This PR adds a feature where the GCS contents manager can run alongside the standard file-based contents manager.

This works by creating two virtual, top-level directories, `GCS` and `Local Disk`. The `GCS` directory maps to the GCS contents manager, and the `Local Disk` directory maps to the file-based contents manager.

That works well enough for reading and writing files from either GCS or local disk, but there is a subtle issue that can come up with JupyterLab if you use any widgets that use the FileBrowser's current path (such as the jupyterlab_git extension).

Those extensions might assume that the API paths being used map to an actual path on disk relative to the content manager's `root_dir` property.

To accommodate this, the `CombinedContentsManager` class has its own `root_dir` property that is separate from the `FileContentsManager` class. User's can set the `c.FileContentsManager.root_dir` config entry to their real notebooks directory, and then create a symlink somewhere on their disk to that directory named `Local Disk` and make the parent directory of the symlink the value of the `c.CombinedContentsManager.root_dir` config entry.

This change fixes #1

### DIFF
--- a/gcs_contents_manager.py
+++ b/gcs_contents_manager.py
@@ -65,7 +65,7 @@ from notebook.services.contents.filemanager import FileContentsManager
 from notebook.services.contents.manager import ContentsManager
 from notebook.services.contents.checkpoints import Checkpoints, GenericCheckpointsMixin
 from tornado.web import HTTPError
-from traitlets import Unicode, default, observe
+from traitlets import Unicode, default
 
 from google.cloud import storage
 
--- a/gcs_contents_manager.py
+++ b/gcs_contents_manager.py
@@ -24,27 +24,48 @@
 # Usage: Add the following lines to your Jupyter config file
 # (e.g. jupyter_notebook_config.py):
 #
-#   from gcs_contents_manager import GCSContentsManager
-#   c.NotebookApp.contents_manager_class = GCSContentsManager
+#   from gcs_contents_manager import CombinedContentsManager, GCSContentsManager
+#   c.NotebookApp.contents_manager_class = CombinedContentsManager
 #   c.GCSContentsManager.bucket_name = '${NOTEBOOK_BUCKET}'
 #   c.GCSContentsManager.bucket_notebooks_path = '${NOTEBOOK_PATH}'
+#   c.GCSContentsManager.project = '${NOTEBOOK_BUCKET_PROJECT}'
+#   c.FileContentsManager.root_dir = '${LOCAL_DISK_NOTEBOOK_DIR}'
 #
 # For '${NOTEBOOK_BUCKET}' specify the name of the GCS bucket where
 # you want to store your notebooks, and for '${NOTEBOOK_PATH}',
 # specify the name of the directory within that bucket that will be
-# treated as your root directory by Jupyter.
+# treated as your root directory by Jupyter. For
+# '${NOTEBOOK_BUCKET_PROJECT}', specify the ID of the GCP project
+# that owns the GCS bucket.
+#
+# If you run JupyterLab with widgets that assume the current file
+# browser path is a location on your local disk (e.g. the
+# jupyterlab-git extension), then you will also need to set up a
+# link somewhere on your local disk for those widgets to use.
+#
+# For example, you could run the following:
+#
+#   mkdir -p ~/.jupyter/symlinks_for_jupyterlab_widgets
+#   ln -s ${LOCAL_DISK_NOTEBOOK_DIR} ~/.jupyter/symlinks_for_jupyterlab_widgets/Local\ Disk
+#
+# And then add the following snippet to your Jupyter config:
+#
+#   c.CombinedContentsManager.root_dir = '~/.jupyter/symlinks_for_jupyterlab_widgets'
 
 import base64
+import errno
 import json
 import logging
 import mimetypes
 import posixpath
 
 import nbformat
+from notebook.services.contents.filecheckpoints import GenericFileCheckpoints
+from notebook.services.contents.filemanager import FileContentsManager
 from notebook.services.contents.manager import ContentsManager
 from notebook.services.contents.checkpoints import Checkpoints, GenericCheckpointsMixin
 from tornado.web import HTTPError
-from traitlets import Unicode, default
+from traitlets import Unicode, default, observe
 
 from google.cloud import storage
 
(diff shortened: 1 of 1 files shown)


gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+
+  def save(self, model, path):
+    if path in ['', '/']:
+      raise HTTPError(403, 'The top-level directory is read-only')
+    try:
+      self.run_pre_save_hook(model=model, path=path)
+
+      cm, relative_path, path_prefix = self._content_manager(path)
+      if not cm:
+        raise HTTPError(404, 'Not Found')
```
karth295, line 645: Nit: similarly, can this be pushed into `_content_manager`? As far as I can tell, only `file_exists` and `dir_exists` don't need to raise 404

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+      if str(err).startswith('Permission denied'):
+        raise HTTPError(403, str(err))
+      raise HTTPError(500, 'Internal server error: [{}] {}'.format(err.errno, str(err)))
+    except HTTPError as err:
+      raise err
+    except Exception as ex:
+      raise HTTPError(500, 'Internal server error: [{}] {}'.format(type(ex), str(ex)))
+
+  def rename_file(self, old_path, new_path):
+    if old_path in ['', '/']:
```
karth295, line 689: I think this check needs to happen on new_path as well so that you don't return 404 when trying to rename to root. (Or at least I'm not sure which is the intended behavior)

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+        raise HTTPError(403, 'The top-level directory contents are read-only')
+
+      new_cm, new_relative_path, _ = self._content_manager(new_path)
+      if not new_cm:
+        raise HTTPError(404, 'Not Found')
+      if new_relative_path in ['', '/']:
+        raise HTTPError(403, 'The top-level directory contents are read-only')
+
+      if old_cm != new_cm:
+        raise HTTPError(400, 'Bad Request')
```
karth295, line 705: Similarly, this could have a better error message noting that you can't rename across file systems

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+
+    file_cm = FileContentsManager(**kwargs)
+    file_cm.checkpoints = GenericFileCheckpoints(**file_cm.checkpoints_kwargs)
+    gcs_cm = GCSContentsManager(**kwargs)
+    self._content_managers = {
+      'Local Disk': file_cm,
+      'GCS': gcs_cm,
+    }
+
+  def _content_manager(self, path):
```
JerryLeiDing, line 546: Nit: _content_manager_for_path for consistency with corresponding CheckpointManager method above?

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+    checkpoint_manager, relative_path = self._checkpoint_manager_for_path(path)
+    if not checkpoint_manager:
+      raise HTTPError(500, 'Unsupported checkpoint path: {}'.format(path))
+    return checkpoint_manager.checkpoint_blob(
+        checkpoint_id, relative_path, create_if_missing=create_if_missing)
+
+  def create_file_checkpoint(self, content, format, path):
+    checkpoint_manager, relative_path = self._checkpoint_manager_for_path(path)
+    if not checkpoint_manager:
+      raise HTTPError(500, 'Unsupported checkpoint path: {}'.format(path))
```
JerryLeiDing, line 484: Why do these all return 500 when the corresponding error modes for the ContentsManager return 403/404?

gcs_contents_manager.py
JerryLeiDing: How difficult would be to support renames/moves across file systems? It seems like a pretty useful feature to e.g move my local notebooks into GCS and vice versa from the UI.

ojarjur:
> Higher level question: this seems like a super useful utility for Juypter in general. Could this be a separate github repo? And could it delegate to arbitrary ContentsManagers configured at paths.

I think it should support arbitrary nested ContentsManagers eventually.

However, I would not want to move it into its own repo. We could bundle it in a stand-alone package on PyPI, but that doesn't mean it should necessarily be developed in a separate git repository.

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+    return checkpoint_manager.list_checkpoints(relative_path)
+
+  def rename_checkpoint(self, checkpoint_id, old_path, new_path):
+    checkpoint_manager, old_relative_path = self._checkpoint_manager_for_path(old_path)
+    if not checkpoint_manager:
+      raise HTTPError(500, 'Unsupported checkpoint path: {}'.format(old_path))
+    new_checkpoint_manager, new_relative_path = self._checkpoint_manager_for_path(new_path)
+    if new_checkpoint_manager != checkpoint_manager:
+      raise HTTPError(500, 'Unsupported rename: {}->{}'.format(old_path, new_path))
```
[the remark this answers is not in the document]
ojarjur: Done

gcs_contents_manager.py
```diff
@@ -428,3 +449,262 @@ def rename_file(self, old_path, new_path):
...
+    for path_prefix in self._content_managers:
+      if path == path_prefix or path.startswith(path_prefix+'/'):
+        relative_path = path[len(path_prefix):]
+        return self._content_managers[path_prefix].checkpoints, relative_path
+    return None, path
+
+  def checkpoint_path(self, checkpoint_id, path):
+    checkpoint_manager, relative_path = self._checkpoint_manager_for_path(path)
+    if not checkpoint_manager:
+      raise HTTPError(500, 'Unsupported checkpoint path: {}'.format(path))
```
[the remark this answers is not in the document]
ojarjur: Done

gcs_contents_manager.py
ojarjur: Done

gcs_contents_manager.py
ojarjur: @JerryLeiDing It would probably be feasible to support renames across file systems, but that is a large enough effort that I would want to do it in its own pull request, and only do it after we have some form of integration tests.

gcs_contents_manager.py
ojarjur: Changed to a 400 error code instead of 500

gcs_contents_manager.py
ojarjur: Done

gcs_contents_manager.py
ojarjur: Done

gcs_contents_manager.py
ojarjur: Fixed to return a 400 instead

PR #14, opened 2020-06-09
Adds gcp-jupyterlab-shared Python project.

prodonjs:
Adds a shared Python subfolder to be used for common routes and functions available to other extensions.

See [README](https://github.com/prodonjs/jupyter-gcs-contents-manager/blob/add-python-shared/shared/server/README.md) for full description.


ojarjur:
Do these enforce/require the same auth on requests that the rest of Jupyter uses? For example, if a user has to sign in to Jupyter with a password, do they also have to sign in before then can issue requests to these handlers?

I want to make sure that these do not introduce any sort of privilege escalation path, so it would be good to verify that no one can call these new handlers unless they can already do things like execute code cells.

Similarly, can we add some sort of allow-list to these handlers to restrict which metadata paths they can fetch or which APIs they can call?

prodonjs:
> Do these enforce/require the same auth on requests that the rest of Jupyter uses? For example, if a user has to sign in to Jupyter with a password, do they also have to sign in before then can issue requests to these handlers?

For AI Platform Notebooks, these paths will be guarded behind the inverting proxy so the same authentication requirements that protect the core Jupyter routes will guard these as well. By default, it looks like the server is protected with [token authentication](https://jupyter-notebook.readthedocs.io/en/stable/security.html), but that _does not_ seem to be respected. It looks like I need to add the [@web.authenticated](https://github.com/jupyterlab/jupyterlab_server/blob/0d8fb81af12949bd9d8c914587a8cee6e40397e5/jupyterlab_server/handlers.py#L60) decorator to each method to respect this.

> I want to make sure that these do not introduce any sort of privilege escalation path, so it would be good to verify that no one can call these new handlers unless they can already do things like execute code cells.
> 
> Similarly, can we add some sort of allow-list to these handlers to restrict which metadata paths they can fetch or which APIs they can call?

Are you specifically thinking about what information they can retrieve from the local machine or what GCP APIs they can call? I was assuming that in the current model, a user with access to JupyterLab has pretty permissive action to the machine itself since they can easily open a terminal shell from the UI and execute the same commands. Likewise, since they would also have access to the machine's service account credential, I was expecting that any API-access matters would be handled by the IAM roles granted to the credential.

ojarjur:
> > Do these enforce/require the same auth on requests that the rest of Jupyter uses? For example, if a user has to sign in to Jupyter with a password, do they also have to sign in before then can issue requests to these handlers?
> 
> For AI Platform Notebooks, these paths will be guarded behind the inverting proxy so the same authentication requirements that protect the core Jupyter routes will guard these as well. By default, it looks like the server is protected with [token authentication](https://jupyter-notebook.readthedocs.io/en/stable/security.html), but that _does not_ seem to be respected. It looks like I need to add the [@web.authenticated](https://github.com/jupyterlab/jupyterlab_server/blob/0d8fb81af12949bd9d8c914587a8cee6e40397e5/jupyterlab_server/handlers.py#L60) decorator to each method to respect this.
> 
> > I want to make sure that these do not introduce any sort of privilege escalation path, so it would be good to verify that no one can call these new handlers unless they can already do things like execute code cells.
> > Similarly, can we add some sort of allow-list to these handlers to restrict which metadata paths they can fetch or which APIs they can call?
> 
> Are you specifically thinking about what information they can retrieve from the local machine or what GCP APIs they can call? I was assuming that in the current model, a user with access to JupyterLab has pretty permissive action to the machine itself since they can easily open a terminal shell from the UI and execute the same commands. Likewise, since they would also have access to the machine's service account credential, I was expecting that any API-access matters would be handled by the IAM roles granted to the credential.

What I was thinking was specifically that a caller who does *not* have access to the existing Jupyter APIs cannot call these new APIs.

E.G. things like what you fixed with the token authentication not being enforced.

Looking at the way the terminals API handles this, it looks like we should also extend [AuthenticatedHandler](https://github.com/jupyter/notebook/blob/b69f22a1b197acc3d9a6d89f2b7a5d9714511dfd/notebook/base/handlers.py#L56) and then invoke [get_current_user](https://github.com/jupyter/notebook/blob/b69f22a1b197acc3d9a6d89f2b7a5d9714511dfd/notebook/base/handlers.py#L128) and return a 403 error if the result is not truthy.

prodonjs:
Ok, the latest commit changes the approach to use the `get_current_user` method in the [prepare](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.prepare) method of a new base handler. This implements the same behavior of returning a 403 when a user is not authenticated via a login cookie to the JupyterLab server.

PR #26, opened 2020-06-15
Create custom request handler for fetching detached comments using git-appraise

mkalil:


jupyterlab_comments/.eslintignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
prodonjs, line 1: This file can be removed in favor of the one in the root of the repository.

jupyterlab_comments/.eslintrc.js
```diff
@@ -0,0 +1,32 @@
+module.exports = {
```
prodonjs, line 1: This file can be removed in favor of the one in the root of the repository.

jupyterlab_comments/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.bundle.*
```
prodonjs, line 1: This file can be removed in favor of the one in the root of the repository.

jupyterlab_comments/.prettierignore
```diff
@@ -0,0 +1,4 @@
+node_modules
```
prodonjs, line 1: This might be worth moving to the root of the repository. By default I think it ignores node_modules and compiled sources, but it is irritating to get warnings when you save a package.json file.

jupyterlab_comments/.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true
```
prodonjs, line 2: This file can be removed in favor of the one in the root of the repository.

jupyterlab_comments/LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
```
prodonjs, line 1: Is there a specific reason not to use the Apache LICENSE from the root repository?

jupyterlab_comments/jupyterlab_comments/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    print("The server extension has been loaded")
+    web_app = nb_server_app.web_app
+    host_pattern = '.*$'
+    base_url = web_app.settings['base_url']
+    hello_route_pattern = url_path_join(base_url, '/hello')
+    detached_comments_route_pattern = url_path_join(base_url, '/detachedComments')
+    verify_repo_route_pattern = url_path_join(base_url, '/verifyRepo')
+    web_app.add_handlers(host_pattern, [
+        (hello_route_pattern, HelloWorldHandler),
```
prodonjs, line 19: Perhaps drop the `HelloWorldHandler` if it was only used as a sanity check during testing/development.

jupyterlab_comments/package.json
```diff
@@ -0,0 +1,54 @@
...
+  "scripts": { 
+    "build": "tsc",
+    "clean": "rimraf lib tsconfig.tsbuildinfo", 
+    "eslint": "eslint . --ext .ts,.tsx --fix",
+    "eslint:check": "eslint . --ext .ts,.tsx",
+    "prepare": "jlpm run clean && jlpm run build",
+    "watch": "tsc -w"
+  },
+  "dependencies": {
+    "@jupyterlab/application": "^2.0.0"
```
prodonjs, line 36: AI Platform Notebooks uses JupyterLab 1.2.x and that's what we've targeted for the other extensions in this repository. 

@ojarjur - This might be something we need to discuss with the team because the last I was told, there are no immediate plans to move to 2.0.x.

jupyterlab_comments/package.json
```diff
@@ -0,0 +1,54 @@
+{
```
prodonjs, line 1: We've tried to keep the package.json file consistent between extensions so that the `lerna` commands at the top of the repository work properly for all subpackages.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/package.json is a good example. The scripts there serve mostly the same purpose but many can be called from the top of the repo to run a build or a test in parallel for each subpackage.

You can then add this subfolder to https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/lerna.json to have it be included.

jupyterlab_comments/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
prodonjs, line 1: I recommend referencing the top-level tsconfig file like https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/tsconfig.json.

We also have a tsconfig.prod.json in each package to exclude any test sources from the production build.

jupyterlab_comments/.prettierignore
mkalil: This file was from an old cookie cutter template, so I think I'm just going to delete it

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class Git(Configurable):
+    """
+    Remote repository should be configured by the
+    user in their Jupyter config file. Default remote is 'origin'
+    """
+    remote = Unicode(u'origin', config=True)
+
+    def run(self, *args):
```
ojarjur, line 14: This needs to somehow get a directory that the user is navigating inside of to pass as the `cwd` keyword argument to the `subprocess.check_output` method.

That could either be an additional argument to this method, or a field of the `Git` class (if you wanted to change this class to no longer be a singleton).

Otherwise, all git commands just run in the same directory where Jupyter was started.

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def status(self):
+        status = self.run('status').decode('utf-8')
+        return status
+
+    def appraise_pull(self):
+        self.run('appraise', 'pull', self.remote)
+
+    # Return true if the current directory is a git repository
+    def inside_git_repo(self):
```
ojarjur, line 29: Right now, I believe this just tells you whether or not Jupyter was started from within a git repository, not whether or not the user is navigating within a git repository.

It needs to set the `cwd` keyword argument similarly to the `run` method above.

jupyterlab_comments/README.md
```diff
@@ -0,0 +1,59 @@
+# jupyterlab_comments
+
+Collaborative notebooks in git
```
ojarjur, line 3: I would recommend using a more specific description here.

Perhaps something like "Notebook comments in git".

jupyterlab_comments/jupyterlab_comments/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+from jupyterlab_comments.handlers import DetachedCommentsHandler, VerifyInsideRepoHandler
+
+def load_jupyter_server_extension(nb_server_app):
+    """
+    Called when the extension is loaded.
+
+    Args:
+        nb_server_app (NotebookWebApplication): handle to the Notebook webserver instance.
+    """
+    print("The server extension has been loaded")
```
ojarjur, line 25: I would recommend making this more specific.

e.g. "The jupyterlab-comments extension has been loaded"

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -0,0 +1,87 @@
...
+        status = self.run(current_path, 'status').decode('utf-8')
+        return status
+
+    def appraise_pull(self, current_path):
+        self.run(current_path, 'appraise', 'pull', self.remote)
+
+    # Return true if the current directory is a git repository
+    def inside_git_repo(self, current_path):
+        try:
+            return_code = subprocess.check_call(['git', 'rev-parse'], cwd=current_path)
```
gogasca, line 47: nit: print return code for user information

PR #28, opened 2020-06-17
feat(optimizer): add optimizer extension from cookiecutter

kpfromer:
This PR adds the files needed for the optimizer extension team to get started. The files were generated from the [cookie cutter project](https://github.com/gclouduniverse/jupyterlab_cookiecutter).


jupyterlab_caip_optimizer/README.md
```diff
@@ -0,0 +1,61 @@
+# Google Cloud Platform AI Optimizer Extension
```
prodonjs, line 1: A lot of these instructions are no longer applicable (probably still need to update part of the Cookie Cutter project). I would suggest trimming it down to include only the things pertinent to your project similar to https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/README.md.

jupyterlab_caip_optimizer/cloudbuild-release.yaml
```diff
@@ -0,0 +1,33 @@
+substitutions:
```
prodonjs, line 1: I think we're going to implement CI/CD with GitHub actions, so you can remove this if you like. Most users won't have permissions to use the `deeplearning-platform-ui-public` bucket anyways.

jupyterlab_caip_optimizer/package.json
```diff
@@ -0,0 +1,115 @@
+{
```
prodonjs, line 1: It doesn't look like the cookie cutter has the latest version of the `package.json`.

Off-hand, I see differences in the `license` and `scripts` sections from the other extensions like https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/package.json.

jupyterlab_caip_optimizer/setup.cfg
```diff
@@ -0,0 +1,5 @@
+[bdist_wheel]
+universal=1
+
+[metadata]
```
prodonjs, line 4: This setting can be removed akin to https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/setup.cfg.

jupyterlab_caip_optimizer/setup.py
```diff
@@ -0,0 +1,183 @@
+from __future__ import print_function
```
prodonjs, line 1: I'd scrap this and base it on the `setup.py` files from any of the existing extension folders. This setup borrowed from core JupyterLab and was overly complex to suit what we needed to do with it.

jupyterlab_caip_optimizer/tests.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
```
prodonjs, line 1: I recently removed these files from each subpackage in favor of a `test-python` command which calls a script hosted in the root of the repo.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/39432248b5a82751c6f8de386177b7f74a625e38/jupyterlab_gcpscheduler/package.json#L35

jupyterlab_caip_optimizer/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
prodonjs, line 1: Prefer referencing the config at the repository root.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/tsconfig.json

PR #29, opened 2020-06-18
Move transport.ts to shared util package.

angelakuo:
Looked into moving ProjectState over as well, but a lot of the methods were scheduler-specific. I think I'll create a separate PR for that because it could get more complicated


shared/client/package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
```
prodonjs, line 31: I would run `npm version patch` to bump the version for the `shared/client` package. You should probably increment the `dependency` entry in scheduler as well.

Before we get ready to release a new Scheduler version, we'll need to make sure the `shared/client` package gets updated on NPM.

PR #42, opened 2020-06-22
Modify shared package

### LINKED ISSUE #31: Remove proxy handlers from gcpscheduler
Now that transport service has been moved to the shared client, we should update the proxy handlers

### LINKED ISSUE #38: Merge shared/client and shared/server into a single package
The code introduced by #29 adds the `ServerProxyTransportService`. This is inherently dependent on the `ProxyHandler` inside of `shared/server/gcp_jupyterlab_shared/handlers.py`.

This package should be refactored to be a single extension package that includes both the Python and client-side code. It will then need to be listed as a Python dependency in the `install_requires` section of each other extension.

angelakuo:
- Renamed gcp-jupyterlab-shared to gcp_jupyterlab_shared
- Merged the 2 subdirectories in shared/
- Scheduler extension now uses shared handler

I plan on updating the README.md but I'm not 100% sure what it should say yet

Resolves #38, resolves #31


package.json
```diff
@@ -41,5 +41,8 @@
       "pre-commit": "npm run lint",
       "pre-push": "npm run lint && npm test"
     }
+  },
+  "dependencies": {
+    "gcp-jupyterlab-shared": "^1.0.0"
```
prodonjs, line 46: I'm not sure why this is getting hoisted up to here. Is there another place where it's being referenced in one of the subpackages possibly?

shared/gcp_jupyterlab_shared/handlers.py
```diff
@@ -189,3 +189,31 @@ async def post(self, base64_url):
   async def put(self, base64_url):
     """Proxies the HTTP PUT request."""
     await self._make_request(base64_url, 'PUT', self.request.body)
+
+class ProjectHandler(APIHandler):
```
prodonjs, line 193: You'll want to make sure these routes get added to https://github.com/angelakuo/jupyter-extensions/blob/721ed550bfb8c32b07d91465ce7f91b468fa140a/shared/gcp_jupyterlab_shared/__init__.py#L40

package.json
angelakuo: Removed it from here and the package-lock.json file, that seems to have worked

shared/gcp_jupyterlab_shared/handlers.py
angelakuo: Oh good catch!

prodonjs:
@angelakuo Looks like there are some conflicts now. You can do a `git pull --rebase upstream master` to replay these changes on top of the current master.

PR #43, opened 2020-06-22
Move AutoML Extension to main repo

shunr:
This PR moves all our existing code from a private repo set up with the cookie cutter. This is a simple move and doesn't change anything outside of the jupyter_automl module other than the `lerna.json`. Subsequent PRs will be made to move our shared components to the shared components module.


jupyterlab_automl/LICENSE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
prodonjs, line 1: This can be dropped in favor of the LICENSE file at the top-level.

jupyterlab_automl/README.md
```diff
@@ -0,0 +1,61 @@
+# Notebook AutoML Extension
```
prodonjs, line 1: This can eventually be cleaned up since the setup instructions should now mirror those for the other extensions.

jupyterlab_automl/cloudbuild-release.yaml
```diff
@@ -0,0 +1,33 @@
+substitutions:
```
prodonjs, line 1: Feel free to drop this as well since are moving to GitHub actions for CI/CD.

jupyterlab_automl/LICENSE
josie-garza: Sure thing.

jupyterlab_automl/cloudbuild-release.yaml
josie-garza: Okay sounds good.

PR #49, opened 2020-06-23
Feat/resource utilization graph component

l1ghtspeed:
Slim wrapper of AreaChart component from recharts, as a functional component that is flexible in taking in various props.

### DIFF
--- a/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
+++ b/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import { AreaChart, Area, Tooltip, CartesianGrid, YAxis } from 'recharts';
+
+type ChartProps = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data: any[];
+  dataKey: string;
+  height: number;
+  width: number;
+  stroke: string;
+  fill: string;
+  hideYAxis: boolean;
+  yDomain: number[];
+  horizontalPoints: number[];
+  showHorizontalGrid: boolean;
+  showVerticalGrid: boolean;
+};
+
+export const AreaChartWrapper = ({
+  data,
+  dataKey,
+  height,
+  width,
+  stroke,
+  fill,
+  hideYAxis,
+  yDomain,
+  horizontalPoints,
+  showHorizontalGrid,
+  showVerticalGrid,
+}: ChartProps) => (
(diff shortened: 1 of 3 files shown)


jupyterlab_gcedetails/src/components/chart_wrapper.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { AreaChart, Area, Tooltip, CartesianGrid, YAxis } from 'recharts';
+
+type ChartProps = {
```
prodonjs, line 4: By convention we've used the `interface` keyword as opposed to `type` for defining object types. Is there a reason for choosing the latter?

jupyterlab_gcedetails/src/components/chart_wrapper.tsx
```diff
@@ -0,0 +1,35 @@
...
+  dataKey: string;
+  title: string;
+  titleClass: string;
+  areaChartProps: {};
+  areaProps: {};
+  cartesianGridProps: {};
+  yAxisProps: {};
+};
+
+export const AreaChartWrapper = ({
```
prodonjs, line 16: We've used the convention of using the `function` keyword to declare functional components (see https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/c1fb81d9c22575255dd372991aeda5df0321aed0/jupyterlab_gcedetails/src/components/details_dialog_body.tsx#L25) as that fits our internal TS style guide.

Also, I think you should be able to just say `props: ChartProps` for the parameter interface.

l1ghtspeed: Ok cool, is it better for the object destructuring to happen in the params or in the function body, or just use `props.xxx` ?

jupyterlab_gcedetails/src/components/chart_wrapper.tsx
l1ghtspeed: No particular reason, will update

jupyterlab_gcedetails/src/components/chart_wrapper.tsx
prodonjs: There's no official guidance in our internal TS style guide, so I'm OK with the destructuring here instead of in the function body.

l1ghtspeed: Ok, I've just moved it to the function body to go in line with `DetailsDialogBody`

PR #50, opened 2020-06-23
Add jupyterlab_bigquery extension

hwing777:
Initial PR to add the BigQuery extension to the main repo

### DIFF
--- a/jupyterlab_caip_optimizer/src/store/store.ts
+++ b/jupyterlab_caip_optimizer/src/store/store.ts
@@ -1,0 +1,16 @@
+import { combineReducers, Reducer } from 'redux';
+import { configureStore, ThunkAction, Action } from '@reduxjs/toolkit';
+import { useDispatch } from 'react-redux';
+
+const rootReducer: Reducer = combineReducers({});
+
+export const store = configureStore({
+  reducer: rootReducer,
+});
+
+export type RootState = ReturnType<typeof rootReducer>;
+export type AppDispatch = typeof store.dispatch;
+export type AppThunk = ThunkAction<void, RootState, unknown, Action<string>>;
+
+// A reusable hook that has resolved types
+export const useAppDispatch = () => useDispatch<AppDispatch>();


Pipfile
```diff
@@ -8,6 +8,7 @@ python_version = "3.7"
 
 [packages]
 jupyterlab = "==1.2.15"
+google-cloud-bigquery = "==1.25.0"
```
prodonjs, line 11: You don't need to change this here. The Pipfile is responsible for setting up the overall requirements for development in any of the packages. In your specific package folder, your extensions `install_requires` section in the `setup.py` file dictates what your particular extension's dependencies are.

jupyterlab_bigquery/.eslintrc.js
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const esLintConfig = require('../.eslintrc');
+
+esLintConfig.rules = {
```
prodonjs, line 18: This is fine to include for now, but where possible we should use the same linting configuration and coding style across all of the subpackages.

jupyterlab_bigquery/cloudbuild-release.yaml
```diff
@@ -0,0 +1,33 @@
+substitutions:
```
prodonjs, line 1: This can be removed as we won't be using Cloud Build to release these extensions.

jupyterlab_bigquery/.eslintrc.js
hwing777: What linting config/coding style are the other subpackages following currently? Is there a unified way or has it yet to be determined?

prodonjs: Basically the rules defined at the repo root should be pretty standard and are in compliance with Google TS style rules for the most part. I think the GCS extension had to deviate from these in a few places because it was forked from the base filebrowser extension.

jupyterlab_gcedetails/README.md
```diff
@@ -13,7 +13,8 @@ In order to access the metadata server running on GCE VM, you will need to
 issue the following command to set up port forwarding. This will allow
 requests made locally to port 8889 to be forwarded to a real GCE metadata
 server. Do this in another terminal window since it will create an interactive
-SSH session.
+SSH session. Replace the instance variable in the URL below with the name of an instance
```
CynthiaXJia, line 16: Confused about why changes that exist in the upstream are appearing as additions from our fork.

PR #51, opened 2020-06-23
[AutoML] Refactor backend handlers

shunr:
- Separated out the request handler logic from the AutoML code, added decorators to simplify the process of making new request handlers 
- Also fixed a display bug in the frontend models display

### DIFF
--- a/jupyterlab_automl/jupyterlab_automl/service.py
+++ b/jupyterlab_automl/jupyterlab_automl/service.py
@@ -1,0 +1,283 @@
+"""Services for AutoML extension backend"""
+
+import math
+from collections import defaultdict
+from enum import Enum
+
+from google.cloud import automl_v1beta1
+
+from gcp_jupyterlab_shared.handlers import AuthProvider
+
+
+def parse_dataset_type(dataset):
+    for dt in DatasetType:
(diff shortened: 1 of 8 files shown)


jupyterlab_automl/jupyterlab_automl/service.py
```diff
@@ -0,0 +1,283 @@
...
+                "weight_column_spec_id": dataset.tables_dataset_metadata.weight_column_spec_id,
+                "ml_use_column_spec_id": dataset.tables_dataset_metadata.ml_use_column_spec_id,
+                "stats_update_time": dataset.tables_dataset_metadata.stats_update_time.ToMilliseconds(),
+            }
+        elif dataset_type == DatasetType.ICN:
+            metadata = {
+                "classification_type": dataset.image_classification_dataset_metadata.classification_type,
+            }
+        elif dataset_type == DatasetType.IOD:
+            metadata = ""
```
josie-garza, line 137: should there be an else here if the dataset is not one of the three types? or you could define metadata outside of the if/else clause.

shunr: +1 will address this

PR #52, opened 2020-06-23
Fetch detached comments for a file

mkalil:
Add ability to get the currently open file and fetch detached comments for that file

### DIFF
--- a/jupyterlab_comments/jupyterlab_comments/git_commands.py
+++ b/jupyterlab_comments/jupyterlab_comments/git_commands.py
@@ -17,72 +17,71 @@ import json
 from traitlets.config.configurable import Configurable
 from traitlets import Int, Float, Unicode, Bool
 
-
 class Git(Configurable):
-    """
-    Remote repository should be configured by the
-    user in their Jupyter config file. Default remote is 'origin'
-    """
-    remote = Unicode(u'origin', config=True)
+	"""
+	Remote repository should be configured by the
+	user in their Jupyter config file. Default remote is 'origin'
+	"""
+	remote = Unicode(u'origin', config=True)
 
-    def __init__(self, **kwargs):
-        super(Git, self).__init__(**kwargs)
+	def __init__(self, **kwargs):
+		super(Git, self).__init__(**kwargs)
 
-    def run(self, current_path, *args):
-        try:
-          return subprocess.check_output(['git'] + list(args), cwd=current_path)
-        except subprocess.CalledProcessError as e:
-          print("Error invoking git command")
-          print(e.output)
+	def run(self, cwd, *args):
+		try:
+		  return subprocess.check_output(['git'] + list(args), cwd=cwd)
+		except subprocess.CalledProcessError as e:
+		  print("Error invoking git command")
+		  print(e.output)
 
-    def status(self, current_path):
-        status = self.run(current_path, 'status').decode('utf-8')
-        return status
+	def status(self, repo_path):
+		status = self.run(repo_path, 'status').decode('utf-8')
+		return status
 
-    def appraise_pull(self, current_path):
-        self.run(current_path, 'appraise', 'pull', self.remote)
+	def appraise_pull(self, repo_path):
+		self.run(repo_path, 'appraise', 'pull', self.remote)
 
-    def inside_git_repo(self, current_path):
-        """
-        Return true if the current directory is a git repository.
-        """
-        try:
-          return_code = subprocess.check_call(['git', 'rev-parse'],
-                                              cwd=current_path)
-          print(return_code)
-          return return_code == 0
-        except Exception as e:
-          print("Error invoking git command")
-          print(e.output)
+	def inside_git_repo(self, path_to_file):
+		"""
+		Return true if the path is inside a git repository.
+		"""
+		try:
+		  return_code = subprocess.call(['git', 'rev-parse'],
+											  cwd=path_to_file)
+		  return return_code == 0
+		except Exception as e:
+		  print("Unexpected error when checking 'git rev-parse' command return code")
+		  print(e)
+
+	def get_repo_root(self, full_file_path):
+		git_repo_root = self.run(full_file_path, 'rev-parse', '--show-toplevel').decode('utf-8').rstrip("\n")
+		return git_repo_root
 
 
-    def get_comments_for_path(self, file_path, current_path):
-        """
-        Returns a JSON list of commments for the given file_path.
+	def get_comments_for_path(self, file_path_from_repo_root, git_root_dir):
+		"""
+		Returns a JSON list of commments for the given file.
 
-        Keys of objects returned: timestamp, author, location, description
-        """
+		Assumes that git_root_dir is a valid path within a Git repo.
+		Keys of objects returned: timestamp, author, location, description
+		"""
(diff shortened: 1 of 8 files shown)


jupyterlab_comments/src/index.ts
```diff
@@ -14,20 +14,117 @@
...
+  Message
+} from '@phosphor/messaging';
+
+import { MainAreaWidget, ICommandPalette } from '@jupyterlab/apputils';
+
+import { PageConfig } from '@jupyterlab/coreutils';
+
+import { httpGitRequest } from './git'
+
+class File {
```
mkalil, line 42: I plan to refactor the File and CommentsWidget classes to mimic the new cookie cutter, in another commit

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -39,6 +41,15 @@ class DetachedCommentsHandler(APIHandler):
   def get(self):
     file_path = self.get_argument('file_path')
     current_path = self.get_argument('current_path')
+
+    if current_path.startswith("~"):
```
gogasca, line 45: is it always guaranteed that file_path and current_path return a value? otherwise you may want to do raise exceptions.

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -16,8 +16,10 @@
 from jupyterlab_comments.git_commands import Git
 import traitlets.config
 import json
-git = Git(config=traitlets.config.get_config()
-         )  #global instance of connection to git commands
+from pathlib import Path
+
+home = str(Path.home())
```
gogasca, line 21: Not sure why home is declared outside the get() method of PreviousNamesHandler. Can you just call it directly?

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -47,6 +47,7 @@ def inside_git_repo(self, current_path):
     try:
```
gogasca, line 47: The doc string should be one sentence. You can put additional description in a separate paragraph.

```
def f():
  ""one sentence doc string.

  More description here in a separate paragragh.
  ""
```

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -39,6 +41,15 @@ class DetachedCommentsHandler(APIHandler):
   def get(self):
     file_path = self.get_argument('file_path')
     current_path = self.get_argument('current_path')
+
+    if current_path.startswith("~"):
+        """
+        Replace the '~' with the full path to the user's home directory.
+        The modified path is needed as input to a subprocess call (setting
+        the current working directory)
+        TODO: deal with how other OS represent home directory
```
gogasca, line 50: Normally use TODO(yourusername) to assign a person :)

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -16,46 +16,63 @@
...
+            current_path = self.get_argument('current_path')
+
+            if current_path.startswith("~"):
+                """
+                Replace the '~' with the full path to the user's home directory.
+                The modified path is needed as input to a subprocess call (setting
+                the current working directory)
+                TODO (mkalil): deal with how other OS represent home directory
+                """
+                current_path = "".join([str(Path.home()), current_path[1:]])
```
ojarjur, line 54: I think this can be `current_path = os.path.expanduser(current_path)`

See details [here](https://docs.python.org/3/library/os.path.html#os.path.expanduser)

jupyterlab_comments/src/git.ts
```diff
@@ -0,0 +1,50 @@
...
+import { URLExt } from '@jupyterlab/coreutils';
+
+
+/*
+Send a request to the server
+Params:
+    handler = name of the handler route as defined in the jupyterlab_comments module
+    method = GET or POST
+    fileName = file path relative to the root of the Jupyter Lab server
+    currentPath = path from home directory to root of Jupyter Lab server
```
ojarjur, line 27: This description seems inaccurate. This value doesn't seem like it is always relative to the user's home directory. Instead, it seems to be based on whatever Jupyter was started with (so it may be relative to the user's home directory if the jupyter config was defined to be relative to their home dir).

I would just call this `Path to the root directory of the Jupyter server`.

It might also be worth calling it `serverRoot` to maintain consistency with the other references to it.

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -16,46 +16,63 @@
...
 
-  def get(self):
-    current_path = self.get_argument('current_path')
-    if git.inside_git_repo():
-      self.finish("Current directory is a git repository.")
-    else:
-      self.finish("Current directory is NOT a git repository.")
+    def get(self):
+        current_path = self.get_argument('current_path')
+        if git.inside_git_repo():
```
ojarjur, line 75: The path needs to be passed to the `inside_git_repo` method.

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -16,46 +16,63 @@
...
 class VerifyInsideRepoHandler(APIHandler):
 
-  def get(self):
-    current_path = self.get_argument('current_path')
-    if git.inside_git_repo():
-      self.finish("Current directory is a git repository.")
-    else:
-      self.finish("Current directory is NOT a git repository.")
+    def get(self):
+        current_path = self.get_argument('current_path')
```
ojarjur, line 74: This assumes the jupyter root directory is a git repository, but it might not be. Instead, the root dir may contain subdirectories that are git repos, and we should support that scenario.

I think the path needs to be constructed by doing something like `os.path.dirname(os.path.join(self.get_argument('current_path'), self.get_argument('file_path')))`.

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -19,72 +19,70 @@
...
-        2. Use 'git appraise show' to get the comments JSON object
-        3. Parse and return comments
+        Keys of objects returned: timestamp, author, location, description
         """
-    pass
 
-  def add_comment(self, file_path, current_path):
-    pass
+        if self.inside_git_repo(current_path):
+          #self.appraise_pull(current_path) #pull new comments from remote repo
```
ojarjur, line 67: Sorry, I am just now realizing that the path you use here cannot be the same `file_path` you get from JupyterLab.

Instead, it needs to be the relative path to the file from the root directory of the git repository (which might be a sub-directory of the Jupyter root dir).

You can find the root directory of the git repository using the command `git rev-parse --show-toplevel`.

You can then make the file_path relative to that using `os.path.relpath(file_path, start=git_root_dir)`, where `git_root_dir` is the output of the `git rev-parse...` command above.

jupyterlab_comments/jupyterlab_comments/handlers.py
mkalil: I think I'm actually just going to delete the VerifyInsideRepo handler since I haven't been using it. Instead I'm calling `self.inside_git_repo(current_path))` from the `get_comments_for_path` method in git_commands.py

PR #53, opened 2020-06-24
(WIP) Feat/resource util data retrieval

l1ghtspeed:
Just wanted to get some quick opinions, ignore the chart_wrapper, package.json, and readme changes are covered in #49

### DIFF
--- a/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
+++ b/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
@@ -1,0 +1,47 @@
+import * as React from 'react';
+import { AreaChart, Area, Tooltip, CartesianGrid, YAxis } from 'recharts';
+
+type ChartProps = {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data: any[];
+  dataKey: string;
+  height: number;
+  width: number;
+  stroke: string;
+  fill: string;
+  hideYAxis: boolean;
+  yDomain: number[];
+  horizontalPoints: number[];
+  showHorizontalGrid: boolean;
+  showVerticalGrid: boolean;
+};
+
+export const AreaChartWrapper = ({
+  data,
+  dataKey,
+  height,
+  width,
+  stroke,
+  fill,
+  hideYAxis,
+  yDomain,
+  horizontalPoints,
+  showHorizontalGrid,
+  showVerticalGrid,
+}: ChartProps) => (
+  <AreaChart height={height} width={width} data={data}>
+    <Area
+      stroke={stroke}
+      fill={fill}
+      isAnimationActive={false}
+      dataKey={dataKey}
+    />
+    <Tooltip />
+    <CartesianGrid
+      horizontalPoints={horizontalPoints}
+      horizontal={showHorizontalGrid}
+      vertical={showVerticalGrid}
+    />
+    <YAxis hide={hideYAxis} domain={yDomain} />
+  </AreaChart>
+);
--- a/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
+++ b/jupyterlab_gcedetails/src/components/chart_wrapper.tsx
@@ -7,15 +7,10 @@ type ChartProps = {
   dataKey: string;
   title: string;
   titleClass: string;
-  height: number;
-  width: number;
-  stroke: string;
-  fill: string;
-  hideYAxis: boolean;
-  yDomain: number[];
-  horizontalPoints: number[];
-  showHorizontalGrid: boolean;
-  showVerticalGrid: boolean;
+  areaChartProps: {};
+  areaProps: {};
+  cartesianGridProps: {};
+  yAxisProps: {};
 };
 
 export const AreaChartWrapper = ({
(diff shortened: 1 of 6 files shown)


jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+  STYLES,
+} from '../data';
+
+interface State {
+  data?: Utilization[];
+  receivedError: boolean;
+}
+
+export class ResourceUtilizationCharts extends React.Component<{}, State> {
+  private readonly serverSettings = ServerConnection.defaultSettings;
```
l1ghtspeed, line 18: Not sure if it's overengineering but maybe we can have a server object initialized in a parent component and have that be passed down as a props instead of duplicating this code.

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+        {},
+        this.serverSettings
+      );
+      if (!response.ok) {
+        this.setState({ receivedError: true });
+        return;
+      }
+      const details = await response.json();
+      const data = this.state.data.slice(1);
+      data.push(details.utilization);
```
l1ghtspeed, line 84: Also probably not in the scope of this PR but would be nice to have a separate get handler in the python backend for the static machine details and the dynamic utilization data

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+      this.setState({ data: data });
+    } catch (e) {
+      console.warn('Unable to retrieve GCE VM details');
+      console.log(e);
```
l1ghtspeed, line 88: Will remove this during cleanup

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+            />
+          </span>
+        )}
+      </span>
+    );
+  }
+
+  private async pollUtilizationData() {
+    try {
+      const response = await ServerConnection.makeRequest(
```
l1ghtspeed, line 73: Also probably not in the scope of this PR but would be nice to have a separate get handler in the python backend for the static machine details and the dynamic utilization data

jupyterlab_gcedetails/src/data.ts
```diff
@@ -18,10 +18,10 @@ import * as csstips from 'csstips';
 import { stylesheet } from 'typestyle';
 
 interface MachineType {
-    name: string;
-    description: string;
```
l1ghtspeed, line 22: These are just some spacing changes made by the linter I think

jupyterlab_gcedetails/src/data.ts
```diff
@@ -140,4 +140,32 @@ export const STYLES = stylesheet({
     display: 'table-row',
     boxShadow: 'inset 0 -1px 0 0 var(--jp-border-color0)',
   },
+  chartTitleSmall: {
+    fontSize: '20px',
+    marginLeft: '20px',
+  },
 });
+
+export const AREA_CHART_BLUE = {
```
l1ghtspeed, line 149: Should the styles related consts in this file be moved to the style folder?

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
prodonjs: Yeah I would suggest factoring out the code that interacts with the server-side API to a common module within the extension so it can be shared by both this component and the parent `VmDetails`.

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+  private readonly NUM_DATA_POINTS = 20;
+  private readonly REFRESH_INTERVAL = 1000;
+  constructor(props: {}) {
+    super(props);
+    const data = [];
+    for (let i = 0; i < this.NUM_DATA_POINTS; i++) {
+      data.push({ cpu: 0, memory: 0 });
+    }
+    this.state = {
+      data: data,
```
yaminahq, line 30: nit: you can probably just replace this with data

jupyterlab_gcedetails/src/components/chart_wrapper.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface ChartProps {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  data: any[];
+  dataKey: string;
+  title: string;
+  titleClass: string;
+  areaChartProps: {};
+  areaProps: {};
+  cartesianGridProps: {};
+  yAxisProps: {};
```
yaminahq, line 13: should we have more specific types defined here for our current use or are you trying to make the component more general?

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,91 @@
...
+        this.serverSettings
+      );
+      if (!response.ok) {
+        this.setState({ receivedError: true });
+        return;
+      }
+      const details = await response.json();
+      const data = this.state.data.slice(1);
+      data.push(details.utilization);
+      this.setState({ data: data });
```
prodonjs, line 85: You can use the shortened syntax here if you like `this.setState({ data });`

jupyterlab_gcedetails/src/data.ts
yaminahq: Hmm in general in this repo styles are defined in the typescript files and the styles folder is just used to store and define images but I don't have a strong opinion on this. By the way if the specific constants you defined here only apply to a single component its probably better to just define them in the corresponding components file instead of defining and exporting from here

jupyterlab_gcedetails/src/components/chart_wrapper.tsx
l1ghtspeed: trying to make it more general, but I will update to give the objects more specific shape in this PR since the props are pretty well defined from the recharts api

l1ghtspeed: my thinking is that eventually if this becomes a shared component we would have different base configurations of the props set so everything is stylistically similar, and people can override them as needed.

jupyterlab_gcedetails/src/data.ts
l1ghtspeed: Yea true, I was thinking of having the color ones here in case they're needed again, but it might be better to export just the hex for the color profiles, unless its already done elsewhere?

yaminahq: If you mean general shared colors profiles across this repo, some are defined here: https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/shared/src/styles.ts#L32 ?

l1ghtspeed:
Closing for updated and non WIP PR on other branch

PR #54, opened 2020-06-24
feature(optimizer): add main area widget and view manager

kpfromer:
# Adds:

- Async slice - a potential tool for using reusable async logic for studies/trials (_This might be removed in favor of [`createAsyncThunk`](https://redux-toolkit.js.org/api/createAsyncThunk)_)
- Added `viewSlice` to the redux store
- `MainAreaWidget` - a widget that is managed by the view state in the redux store, ie it can be closed and opened from any component via redux actions

_Please ignore the bad branch name_ :sweat_smile:


jupyterlab_caip_optimizer/src/components/main_area_widget.tsx
```diff
@@ -0,0 +1,51 @@
...
+const mapStateToProps = (state: RootState) => ({
+  data: state.view.data,
+});
+
+/**
+ * ViewManager acts as the extensions router for various views.
+ */
+export const ViewManager = ({ data }: { data: ViewType }) => {
+  // TODO: add custom components
+  if (data.view === 'dashboard') {
```
angelakuo, line 16: If ViewType gets any larger I'd consider turning this into a switch statement

kpfromer:
The main area is not developed (it will be by Jun in the feature 2 phase) but here's what it looks like at the moment:
dashboard
_it's quite underwhelming_.

jupyterlab_caip_optimizer/src/store/async_slice.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    let initialState: AsyncState<unknown>;
+    beforeEach(() => {
+      initialState = {
+        loading: true,
+        error: null,
+        data: ['data'],
+      };
+    });
+    it('stops loading', () => {
+      const newState = genericSlice.reducer(
```
angelakuo, line 91: Seems like this could be in the beforeEach block?

jupyterlab_caip_optimizer/src/store/async_slice.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+    it('removes error', () => {
+      const newState = genericSlice.reducer(
+        initialState,
+        genericSlice.actions.success('data')
+      );
+      expect(newState.error).toBeNull();
+    });
+  });
+  describe('error', () => {
+    let initialState: AsyncState<unknown>;
```
angelakuo, line 82: Could this be a const since you're not reassigning it?

prodonjs:
Just a general comment. I noticed during a fresh installation there are a few warnings from Lerna about inconsistent versions of libraries used by optimizer versus some of the other extensions. I think most of this is due to the older version of the cookiecutter project that you guys might have started from.

```
erna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on @typescript-eslint/eslint-plugin@^2.17.0, which differs from the hoisted @typescript-eslint/eslint-plugin@^2.24.0.
lerna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on @typescript-eslint/parser@^2.17.0, which differs from the hoisted @typescript-eslint/parser@^2.24.0.
lerna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on eslint@^6.3.0, which differs from the hoisted eslint@^6.8.0.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on eslint-config-prettier@^6.10.0, which differs from the more common eslint-config-prettier@^6.2.0.
lerna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on eslint-config-prettier@^6.2.0, which differs from the hoisted eslint-config-prettier@^6.10.0.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on eslint-plugin-prettier@^3.1.2, which differs from the more common eslint-plugin-prettier@^3.1.0.
lerna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on eslint-plugin-prettier@^3.1.0, which differs from the hoisted eslint-plugin-prettier@^3.1.2.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on prettier@^1.19.1, which differs from the more common prettier@^1.18.2.
lerna WARN EHOIST_PKG_VERSION "jupyterlab_caip_optimizer" package depends on prettier@^1.18.2, which differs from the hoisted prettier@^1.19.1.
...
```

Would that be something you could address either in this PR or in a subsequent one? I don't think any of the changes will make a difference to the project.

kpfromer:
Oh, I did not see that when working on adding the extension. We will fix the packages in a separate PR.

PR #56, opened 2020-06-24
Add BigQuery Extension

CynthiaXJia:
Adding extension in jupyterlab_bigquery


jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -0,0 +1,166 @@
...
+from google.cloud import bigquery
+
+# import google.auth
+# from google.auth.exceptions import GoogleAuthError
+# from google.auth.transport.requests import Request
+
+from jupyterlab_bigquery.version import VERSION
+
+SCOPE = ("https://www.googleapis.com/auth/cloud-platform",)
+client = bigquery.Client()
```
cbwilkes, line 25: I would recommend structuring this file's bigquery client to be mockable for testing. For example look in https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_automl/jupyterlab_automl/handlers.py

Remove the global client. Then create a function to generate the client like:

```
def create_bigquery_client():
    return bigquery.Client()
```

Change these functions to receive the client, so that we can test these in a later PR:
```
list_projects(client, ...)
list_datasets(client, ...)
list_tables(client, ...)
list_models(client, ...)
get_dataset_details(client, ...)
get_table_details(client, ...)
```

Create the client inside of the ApiHandler functions and pass them into the function.
```
class MyHandler(APIHandler):
  bigquery_client = None

  @gen.coroutine
  def get(self, *args, **kwargs):
     try:
       self.bigquery_client = create_bigquery_client()
       list_projects(self.bigquery_client)
       ...
```

jupyterlab_bigquery/jupyterlab_bigquery/tests/handlers_test.py
```diff
@@ -0,0 +1,13 @@
+import unittest
+from unittest.mock import Mock, MagicMock, patch
+
+from jupyterlab_bigquery import handlers
+
+
+class TestList(unittest.TestCase):
```
cbwilkes, line 7: Add a TODO to add tests here for these functions we get a chance:
  - list_projects, list_datasets, list_tables, list_models, get_dataset_details, get_table_details

PR #59, opened 2020-06-26
Adding popup widget component

yaminahq:
Adding a popup widget component that adds an icon to the status bar which opens a popup anchored to the icon when clicked. The popup displays the components children. Also slightly modifying the existing stylizing of the widget to match the onHover behavior of the other widgets on the status bar more closely.


jupyterlab_gcedetails/src/components/widget_popup.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { STYLES } from '../data';
+
+interface Props {
+  children: React.ReactNode;
+}
+
+const GRAPH_ICON_CLASS = 'jp-UtilizationGraphsIcon';
+
+export function WidgetPopup(props: Props) {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
prodonjs, line 29: In general I haven't used `useState` and stuck with the functional component style. In those cases I mostly just stuck with class-based components but I'm perfectly fine with this.

yaminahq: In that case, I'll change this component to be a class-based component to better match the rest of the code then :)

PR #60, opened 2020-06-26
Feat/resource utilization graph component

l1ghtspeed:
Still working on the tests, but wanted to get some more feedback on implementation specifically for the ServerWrapper file.
Previous WIP PR can be found here #53


yaminahq:
LGTM 🚀

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -0,0 +1,102 @@
...
+    super(props);
+    const data = [];
+    for (let i = 0; i < this.NUM_DATA_POINTS; i++) {
+      data.push({ cpu: 0, memory: 0 });
+    }
+    this.state = {
+      data,
+      receivedError: false,
+    };
+    this.refreshInterval = window.setInterval(() => {
```
prodonjs, line 53: I would suggest creating this interval when the component mounts rather than the constructor. In general that seems like the suggested place to initiate any calls to a backend.

https://reactjs.org/docs/react-component.html#componentdidmount

jupyterlab_gcedetails/src/components/server_wrapper.tsx
```diff
@@ -0,0 +1,30 @@
...
+  private readonly url: string;
+  constructor(
+    endpoint: string,
+    serverSettings: ServerConnection.ISettings = ServerConnection.defaultSettings
+  ) {
+    this.serverSettings = serverSettings;
+    this.url = this.serverSettings.baseUrl + endpoint;
+  }
+
+  async get() {
```
prodonjs, line 14: I would suggest making this interface more explicit. What kind of data is actually being returned? Maybe something like `getUtilizationData`. I would also suggest that the return type be `Promise<Utilization>`. Then, you can just behave like a normal promise and reject if the method is unable to fetch data, rather than return the object with the `ok` property.

jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
l1ghtspeed: I think since `refreshInterval`  is readonly type, it can't be assigned in `componentDidMount` which is why I think it's assigned in the constructor in the original implementation in `details_widget.tsx` as well.

prodonjs: That is correct. However in `details_widget.tsx`, the logic created the interval but used the `shouldRefresh` state variable to determine whether the call to the server should be made, which would initially be false. In this case, the visibility of the chart determines that the data should be refreshed. You could drop the `readonly` modifier from the property and assign a new interval in the `componentDidMount`, and clear it in `componentWillUnmount`.

prodonjs:
This looks good. It looks like you just need to clean up some of the test snapshots.

See https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/60/checks?check_run_id=822842298#step:8:126

PR #64, opened 2020-06-26
[Scheduler] Refactors to remove all in-package Python code.

prodonjs:
- Relies on handlers from gcp_jupyterlab_shared
- Introduces new getMetadata function in gcp_jupyterlab_shared
- Retrieves projectId from getMetadata rather than from dedicated project handler
- Allows for a custom machine type to request no GPUs


prodonjs:
@angelakuo - Most of these are just structural changes. The changes to the Scheduler code that matter can be viewed at https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/64/commits/0d2f2d6a5b089f58892e27a9e3db117011d1795d.

PR #80, opened 2020-06-30
[Optimizer] Setup for handling API calls and displaying views

junhyeokkwak:
* Implemented `OptimizerService` that handles API calls (create study, fetch study list, etc.)
* Basic tests for API calls
* Integration of Redux store and `OptimizerService` API handler


jupyterlab_caip_optimizer/src/store/studies.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Study, AsyncState } from '../types';
+import { OptimizerService } from '../service/optimizer';
+import { ServerProxyTransportService } from 'gcp_jupyterlab_shared';
+
+const optimizer = new OptimizerService(new ServerProxyTransportService());
+
+export const fetchStudies = createAsyncThunk<Study[]>(
+  'studies/fetch',
+  async () => {
+    // TODO: use metadata service instead of hardcoded values
```
kpfromer, line 11: Just want to note that we are hardcoding in project id and region data since we were unable to get the metadata server working.

jupyterlab_caip_optimizer/src/service/optimizer.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Class to interact with Optimizer
+ */
+export class OptimizerService {
+  private readonly serverSettings = ServerConnection.defaultSettings;
+
+  constructor(private _transportService: TransportService) {}
+
+  set transportService(transportService: TransportService) {
```
angelakuo, line 42: Are we planning on using this at any point?

jupyterlab_caip_optimizer/src/service/optimizer.ts
```diff
@@ -0,0 +1,86 @@
...
+      const response = await this._transportService.submit<Study>({
+        path: `${ENDPOINT}/projects/${metadata.projectId}/locations/${
+          metadata.region
+        }/studies?study_id=${encodeURI(study.name)}`,
+        method: 'POST',
+        body,
+      });
+      return response.result;
+    } catch (err) {
+      console.error('Unable to create study');
```
angelakuo, line 73: nit: I'd append the err message at the end for easier debugging

jupyterlab_caip_optimizer/src/service/optimizer.ts
```diff
@@ -0,0 +1,86 @@
...
+    });
+    const metadata_response: MetadataRequired = {
+      projectId: response.result.project,
+      region: zoneToRegion(response.result.zone),
+    };
+    return metadata_response;
+  }
+
+  async createStudy(study: Study, metadata: MetadataRequired): Promise<Study> {
+    try {
```
angelakuo, line 61: Should we wrap the other 2 functions in try/catch's?

jupyterlab_caip_optimizer/src/service/test-contsants.ts
```diff
@@ -0,0 +1,73 @@
+/**
```
angelakuo, line 1: File seems to be misspelled (and idk how to make file-level comments)

jupyterlab_caip_optimizer/src/service/test-contsants.ts
```diff
@@ -0,0 +1,73 @@
...
+  StudyConfig,
+  MetricSpec,
+  ParameterSpec,
+  ParameterType,
+  Algorithm,
+  GoalType,
+  State,
+} from '../types';
+
+export const fakeMetric1 = {
```
angelakuo, line 15: I think for all these constants instead of using numbers I would distinguish them further (i.e. fakeMetricUnspecified, fakeParamCategorical, fakeStudyConfigUnspecified) so it's easier to tell what they do

jupyterlab_caip_optimizer/src/service/optimizer.ts
junhyeokkwak: Not really for now, but I assumed it may be useful when we want to change the transport service some time in the future (and GCP Scheduler extension seemed to have this as well). Should I remove it?

jupyterlab_caip_optimizer/src/service/optimizer.ts
junhyeokkwak: I'll fix this

jupyterlab_caip_optimizer/src/service/optimizer.ts
junhyeokkwak: So ideally we'll have different error cases depending on the error codes. Documentation on these different HTTP status codes from Optimizer API is lacking, so we decided to describe it as general error here for now and revisit this afterwards (added in To-do in our project board)

jupyterlab_caip_optimizer/src/service/test-contsants.ts
junhyeokkwak: I'll fix this as well

jupyterlab_caip_optimizer/src/store/studies.ts
angelakuo: I think in that case I would still add functions (e.g. getProjectId and getRegion) to optimizer.ts and return the constants there. That way, when you add in the functionality you only need to edit the functions

jupyterlab_caip_optimizer/src/service/optimizer.ts
angelakuo: It looks like they use it [here](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_gcpscheduler/src/components/dialog.tsx#L274-L281) I generally prefer only adding functions when we actually use them, but if you know that we will definitely be using it in the future I'd say it's fine to keep

jupyterlab_caip_optimizer/src/service/optimizer.ts
angelakuo: Gotcha, I'd leave a TODO comment here as well and link to the documentation for further info

jupyterlab_caip_optimizer/src/service/optimizer.ts
junhyeokkwak: fixed

jupyterlab_caip_optimizer/src/service/optimizer.ts
junhyeokkwak: done

jupyterlab_caip_optimizer/src/service/optimizer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Class to interact with Optimizer
+ */
+export class OptimizerService {
+  private readonly serverSettings = ServerConnection.defaultSettings;
+
+  constructor(private _transportService: TransportService) {}
+
+  /** TODO: Remove hardcoded values and 'option' parameter before release. Currently implemented this way because of ssh restrictions to GCP */
```
angelakuo, line 42: Perhaps we should decide on a line limit for this project/all the projects?

jupyterlab_caip_optimizer/src/service/optimizer.ts
```diff
@@ -0,0 +1,104 @@
...
+  async getMetaData(option: string = 'default-metadata'): Promise<MetadataRequired> {
+    try {
+      if (option === 'default-metadata') {
+        const metadata_default: MetadataRequired = {
```
angelakuo, line 46: I'm seeing some irregularities in case styles (e.g. snake case here, camel case in other files), I think it would be worth it to figure out a consistent case style for this project.

Also somewhat related, figure out when to make consts all caps (like the endpoints const in this file). I believe the convention for our internal source code is that only global constants are capitalized, but definitely feel free to define your own rules

junhyeokkwak: Sounds good. I've tried following those (e.g. capitalizing enums) but I'll try to be more consistent.

PR #86, opened 2020-07-01
[BigQuery] Paged API backend support

MRzNone:
Provides backend support for Paged API

### DIFF
--- a/jupyterlab_bigquery/jupyterlab_bigquery/pagedAPI_handler/pagedAPI_handler.py
+++ b/jupyterlab_bigquery/jupyterlab_bigquery/pagedAPI_handler/pagedAPI_handler.py
@@ -1,0 +1,123 @@
+from enum import Enum
+from notebook.base.handlers import APIHandler, app_log
+from logging import INFO, WARN
+import json
+from threading import Lock
+from collections import deque, defaultdict
+from abc import ABC, abstractmethod
+
+START_STATE = 'start'
+CONTINUE_STATE = 'continue'
+CANCEL_STATE = 'cancel'
+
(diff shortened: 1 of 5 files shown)


jupyterlab_bigquery/jupyterlab_bigquery/pagedAPI_handler/pagedQuery_handler.py
```diff
@@ -0,0 +1,30 @@
+from jupyterlab_bigquery.pagedAPI_handler import PagedAPIHandler
+from google.cloud import bigquery
+import json
+
+class PagedQueryHandler(PagedAPIHandler):
+  def query(self, request_body, page_size):
```
cbwilkes, line 6: Can you write test code for this function?

cbwilkes, line 6: As discussed let's make sure we circle back and test this.

jupyterlab_bigquery/jupyterlab_bigquery/pagedAPI_handler/pagedAPI_handler.py
```diff
@@ -0,0 +1,123 @@
...
+      if query_generator is None:
+          finish = True
+      else:
+          try:
+            load = next(query_generator)
+          except StopIteration:
+            finish = True
+            with PagedAPIHandler.generator_lock:
+                del PagedAPIHandler.generator_pool[id]
+            app_log.log(INFO, 'Successfully finished query %s', id)
```
cbwilkes, line 80: We might should think about clearing out generators after a timeout period, but I don't it is necessary to start.

PR #88, opened 2020-07-06
[HardwareConfiguration] Created hardware scaling form and dialog

### LINKED ISSUE #93: Implement hardware scaling form dialog
Create a form to gather hardware configuration details that will be displayed to the user as a dialog. Based on the Project Beatrix UX Mocks [here](https://docs.google.com/presentation/d/1bPcRicop1_nr2Pxw5mITPyJEiDE8WLq8lyoiDfOTvJw/edit#slide=id.g8514ccd791_4_53).

yaminahq:
Creating UI for the Hardware Scaling form dialog that allows a user to modify the hardware configuration for their notebook.

Screenshot 2020-07-06 at 3 01 54 PM

Screenshot 2020-07-06 at 3 02 09 PM

This commit includes:
- Implementing a SynchronizedSliders component which creates two sliders with text input which can only take on values defined by an array of objects passed as props. The object contains a cpu key which represents the value the first slider can take and a memory key which represent the value the second slider will take consequently and vice versa. This forces the sliders to stay in sync with one another.
- Implementing a HardwareScalingForm component which allows a user to select a machine type, CPU value, memory value and GPU type.
- Implementing a HardwareScalingDialog component which is a dialog thinly wrapping over HardwareScalingForm.
- Adding gcp_jupyterlab shared dependency to gce_details.
- Modifying the shared TextInput component to expose additional input attributes through props.
- Copying the hard-coded Accelerator types data from gcpscheduler for now (will move to a shared location as needed).

Resolves #93

### DIFF
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_dialog.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_dialog.tsx
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Dialog } from '@material-ui/core';
+import * as React from 'react';
+import { HardwareScalingForm } from './hardware_scaling_form';
+
+interface Props {
+  open: boolean;
+  onClose?: () => void;
+}
+
+/** Funtional Component for a common dialog interface with cancel and submit buttons. */
+export function HardwareScalingDialog(props: Props) {
+  return (
+    <Dialog open={props.open}>
+      <HardwareScalingForm onDialogClose={props.onClose} />
+    </Dialog>
+  );
+}
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_dialog.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_dialog.tsx
@@ -20,7 +20,7 @@ import { HardwareScalingForm } from './hardware_scaling_form';
 
 interface Props {
   open: boolean;
-  onClose?: () => void;
+  onClose: () => void;
 }
 
 /** Funtional Component for a common dialog interface with cancel and submit buttons. */
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
@@ -32,6 +32,7 @@ import {
   ACCELERATOR_COUNTS_1_2_4_8,
   machineTypes,
   ACCELERATOR_TYPES,
+  MachineTypeConfigurations,
 } from '../data';
 
 interface Props {
@@ -44,8 +45,8 @@ interface State {
   cpuValue: number;
   memoryValue: number;
   attachGpu: boolean;
-  gpuType?: string;
-  gpuCount?: string;
+  gpuType: string;
+  gpuCount: string;
 }
 
 export const STYLES = stylesheet({
@@ -97,67 +98,48 @@ export class HardwareScalingForm extends React.Component<Props, State> {
     };
   }
 
-  private getMachineType = (baseMachineType: string) => {
+  private getMachineType(baseMachineType: string): MachineTypeConfigurations {
     return machineTypes.find(elem => elem.base.value === baseMachineType);
-  };
+  }
+
+
+  private onAttachGpuChange(event: React.ChangeEvent<HTMLInputElement>) {
+    this.setState({
+      attachGpu: event.target.checked,
+      gpuType: event.target.checked
+        ? (ACCELERATOR_TYPES[1].value as string)
+        : NO_ACCELERATOR,
+      gpuCount: event.target.checked
+        ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
+        : '',
+    });
+  }
+
+  private onBaseMachineTypeChange(event: React.ChangeEvent<HTMLSelectElement>) {
+    this.setState({
+      baseMachineType: event.target.value,
+      cpuValue: this.getMachineType(event.target.value).configurations[0].cpu,
+      memoryValue: this.getMachineType(event.target.value).configurations[0]
+        .memory,
+    });
+  }
+
+  private onGpuTypeChange(event: React.ChangeEvent<HTMLSelectElement>) {
+    this.setState({
+      gpuType: event.target.value,
+      gpuCount: event.target.value
+        ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
+        : '',
+    });
+  }
+
+  // TODO: Implement submit functionality
+  //eslint-disable-next-line @typescript-eslint/no-empty-function
+  private submitForm() {};
 
   render() {
     const { onDialogClose } = this.props;
     const { gpuType, gpuCount, attachGpu, baseMachineType } = this.state;
-
-    const onBaseMachineTypeChange = (
-      event: React.ChangeEvent<HTMLSelectElement>
-    ) => {
-      this.setState({
-        baseMachineType: event.target.value,
-        cpuValue: this.getMachineType(event.target.value).configurations[0].cpu,
-        memoryValue: this.getMachineType(event.target.value).configurations[0]
-          .memory,
-      });
-    };
-
-    const onCpuChange = (newValue: number) => {
-      this.setState({
-        cpuValue: newValue,
-      });
-    };
-
-    const onMemoryChange = (newValue: number) => {
-      this.setState({
-        memoryValue: newValue,
-      });
-    };
-
-    const onGpuTypeChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
-      this.setState({
-        gpuType: event.target.value,
-        gpuCount: event.target.value
-          ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
-          : '',
-      });
-    };
-
-    const onGpuCountChange = (event: React.ChangeEvent<HTMLSelectElement>) => {
-      this.setState({
-        gpuCount: event.target.value,
-      });
-    };
-
-    const onAttachGpuChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-      this.setState({
-        attachGpu: event.target.checked,
-        gpuType: event.target.checked
-          ? (ACCELERATOR_TYPES[1].value as string)
-          : NO_ACCELERATOR,
-        gpuCount: event.target.checked
-          ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
-          : '',
-      });
-    };
-
-    // TODO: Implement submit functionality
-    //eslint-disable-next-line @typescript-eslint/no-empty-function
-    const submitForm = () => {};
 
     return (
       <div className={STYLES.container}>
@@ -171,13 +153,13 @@ export class HardwareScalingForm extends React.Component<Props, State> {
                 name="baseMachineType"
                 value={baseMachineType}
                 options={machineTypes.map(elem => elem.base)}
-                onChange={onBaseMachineTypeChange}
+                onChange={e => this.onBaseMachineTypeChange(e)}
               />
             </div>
             <SynchronizedSliders
               values={this.getMachineType(baseMachineType).configurations}
-              onMemoryChange={onMemoryChange}
-              onCpuChange={onCpuChange}
+              onMemoryChange={memoryValue => this.setState({ memoryValue })}
+              onCpuChange={cpuValue => this.setState({ cpuValue })}
               key={baseMachineType}
             />
             <div className={STYLES.checkboxContainer}>
@@ -186,7 +168,7 @@ export class HardwareScalingForm extends React.Component<Props, State> {
                 className={STYLES.checkbox}
                 name="attachGpu"
                 checked={attachGpu}
-                onChange={onAttachGpuChange}
+                onChange={e => this.onAttachGpuChange(e)}
               />
             </div>
             {attachGpu && (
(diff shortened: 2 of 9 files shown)


jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
```diff
@@ -0,0 +1,242 @@
...
+    );
+  }
+}
+
+interface InputSliderProps {
+  value: number;
+  possibleValues: number[];
+  label?: string;
+  handleSliderChange: (event: any, newValue: number) => void;
+
```
l1ghtspeed, line 195: Weird that this space didn't get picked up by the linter

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,211 @@
...
+interface Props {
+  // Assumes sorted list with memory and cpu set to 0 for first value
+  onDialogClose?: () => void;
+}
+
+interface State {
+  cpuValue: number;
+  memoryValue: number;
+  attachGpu: boolean;
+  gpuType?: string;
```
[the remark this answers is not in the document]
prodonjs: It looks like you've implemented the `gpuType` and `gpuCount` fields similarly to how they work in schedule, so I would just remove the optional annotations here.

As far as making them enums, it really depends on whether they will get passed directly to the service that will actually make the reshaping request. [String enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) may be a good choice, but if we don't find ourselves needing to refer to specific values and instead just pass the selected value into the service method call, I would say that they are generally unnecessary here.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,243 @@
...
+      baseMachineType: machineTypes[0].base.value as string,
+      cpuValue: 0,
+      memoryValue: 0,
+      attachGpu: false,
+      gpuType: NO_ACCELERATOR,
+      gpuCount: '',
+    };
+  }
+
+  private getMachineType = (baseMachineType: string) => {
```
prodonjs, line 100: For consistency, let's use the class-based method definition format rather than lambdas assigned to private class members.

```
private getMachineType(baseMachineType: string): string {
  ...
}
```

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,243 @@
...
+    return machineTypes.find(elem => elem.base.value === baseMachineType);
+  };
+
+  render() {
+    const { onDialogClose } = this.props;
+    const { gpuType, gpuCount, attachGpu, baseMachineType } = this.state;
+
+    const onBaseMachineTypeChange = (
```
prodonjs, line 108: I'd suggest lifting these out to private methods of the class rather than lambdas defined during each call of the `render` method. You may even be able to generalize a method that works for the Cpu, Memory, and GpuCount fields which all take a single value and assign a state property.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,243 @@
...
+          : NO_ACCELERATOR,
+        gpuCount: event.target.checked
+          ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
+          : '',
+      });
+    };
+
+    // TODO: Implement submit functionality
+    //eslint-disable-next-line @typescript-eslint/no-empty-function
+    const submitForm = () => {};
```
prodonjs, line 160: Likewise, I'd suggest lifting this out to a private method and then having the `onClick` property for your submit button call this method.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,243 @@
...
+          />
+        </ActionBar>
+      </div>
+    );
+  }
+}
+
+const DESCRIPTION = `The hardware scaling limits you configured will be the
+max capacity allowed for this notebook. You'll only pay for the hardware 
+resources you use. `;
```
prodonjs, line 232: This description is actually incorrect in our current implementation. The user will be for the time the machine is on.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -0,0 +1,243 @@
...
+const LINK = 'https://cloud.google.com/ai-platform/training/pricing';
```
prodonjs, line 233: This is probably a better link.

https://cloud.google.com/compute/all-pricing

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
```diff
@@ -0,0 +1,246 @@
...
+  /* Assumes sorted list with memory and cpu set to 0 for first value */
+  values: HardwareConfiguration[];
+  onCpuChange: (value: number) => void;
+  onMemoryChange: (value: number) => void;
+}
+
+interface State {
+  memoryValue: number;
+  cpuValue: number;
+  memoryInputDisplayValue: string;
```
prodonjs, line 34: Do you really need to store the string representations of your values for display? It seems like they could easily be inferred in your render method.

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+  render() {
+    const {
+      cpuInputDisplayValue,
+      memoryInputDisplayValue,
+      cpuValue,
+      memoryValue,
+    } = this.state;
+
+    const handleCpuSliderChange = (event: any, newValue: number) => {
```
prodonjs, line 123: As with the other component, I would suggest lifting these to private methods within the class.

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
```diff
@@ -0,0 +1,246 @@
...
+  label?: string;
+  handleSliderChange: (event: any, newValue: number) => void;
+
+  inputDisplayValue: string;
+  handleInputChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  handleInputBlur: () => void;
+  helperText?: string;
+}
+
+function InputSlider(props: InputSliderProps) {
```
prodonjs, line 206: I could see this eventually being useful as a shared component, but I don't think you need to do that now. Still, it might be worth lifting it out to its own file.

jupyterlab_gcedetails/src/data.ts
```diff
@@ -105,6 +111,110 @@ export const REFRESHABLE_MAPPED_ATTRIBUTES = [
...
+  { value: 'n1-highcpu-96', text: '96 CPUs, 86.4 GB RAM' },
+];
+
+export interface HardwareConfiguration {
+  cpu: number;
+  memory: number;
+}
+
+/* CPU to Memory mappings for the Compute Engine machine types */
+interface CEMachineType {
```
prodonjs, line 170: Can we just name this MachineType?

shared/src/components/text_input.tsx
```diff
@@ -24,12 +24,14 @@ interface TextInputProps {
...
   min?: string;
   max?: string;
+  step?: number;
   label?: string;
   name?: string;
   value?: string;
   placeholder?: string;
   hasError?: boolean;
   onChange?: (e: React.ChangeEvent<HTMLInputElement>) => void;
+  onBlur?: () => void;
```
prodonjs, line 34: Perhaps you can follow the same pattern as we used for the `CheckboxInput` and make the props extend `React.InputHTMLAttributes<HTMLInputElement>`. Then, any props that aren't specifically defined for our custom use just get passed down to the <input> element.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/38aefa64465c6073844682fb1297b37d0cf33205/shared/src/components/checkbox_input.tsx#L22

jupyterlab_gcedetails/src/components/hardware_scaling_dialog.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+import { Dialog } from '@material-ui/core';
+import * as React from 'react';
+import { HardwareScalingForm } from './hardware_scaling_form';
+
+interface Props {
+  open: boolean;
+  onClose?: () => void;
```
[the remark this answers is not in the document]
yaminahq: That's a good point, I'll change it to be required

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
```diff
@@ -0,0 +1,242 @@
...
+      <div>
+        <InputSlider
+          possibleValues={this.cpuValues}
+          label="Cores"
+          value={cpuValue}
+          inputDisplayValue={cpuInputDisplayValue}
+          handleSliderChange={handleCpuSliderChange}
+          handleInputChange={handleCpuInputChange}
+          handleInputBlur={handleCpuInputBlur}
+          helperText="1 - 96 vCPUs"
```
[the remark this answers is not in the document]
yaminahq: I ended up changing this to a string interpolation statement as with my latest commit the value for the ranges changes based on the props

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
yaminahq: Good catch!

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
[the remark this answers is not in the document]
yaminahq: Yeah I was trying to follow the gcp scheduler extension closely, I'll remove the optional annotation

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
yaminahq: That's a good point I'll remove the lambdas.
For the generalized method, do you mean something like this:
`  private onChange(key: string, newValue: number) {
    this.setState({
      [String(key)]: newValue,
    } as unknown);
  }`
Or could it also make sense to call setState directly in the onChange props for each of the components in the render method, like [this](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/c1fb81d9c22575255dd372991aeda5df0321aed0/jupyterlab_gcpscheduler/src/components/schedule_builder/hour_schedule_form.tsx#L90)? Not sure of best practice here :/

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
yaminahq: How about this for the description: "The hardware scaling limits you configured will be the
max capacity allowed for this notebook. You'll only pay for the time the 
hardware resources are on."?

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
yaminahq: If I remove the inputDisplayValue variables then the slider automatically moves to different positions while the user is still typing into the input field (jumps from 2 to 20 while the user is entering 20 for example). I thought it would be a better user experience if I only change the value of the slider once the user has finished entering input and the input field loses focus. What do you think?

prodonjs: Ahh I see. So if you calculate the values in the `render` method, they're always going to change as the user changes. This feels similar to what we discussed this morning in that we should try to debounce the change events for the field rather than change state on each value change. After a sufficient amount of time has passed (maybe 1s?), we can consider the user's edit to be complete.

jupyterlab_gcedetails/src/data.ts
yaminahq: MachineType is being used in the file to define the machine the user currently has:
```
interface MachineType {
  name: string;
  description: string;
}

interface Instance {
  attributes: {
    framework: string;
    title: string;
    version: string;
  };
  cpuPlatform: string;
  id: number;
  image: string;
  machineType: MachineType;
  name: string;
  zone: string;
}
```
Thoughts on changing CEMachineType to MachineTypeConfigurations instead?

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
prodonjs: I think either option is reasonable. I think the latter option looks a bit more TS-friendly compared to having to cast the type as `unknown`.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
prodonjs: Yeah I think that's reasonable and we can run the wording by UX design/writing at a later time.

jupyterlab_gcedetails/src/components/synchronized_sliders.tsx
yaminahq: I played around with implementing the debounced handler but doing so also debounces displaying the value the user has typed in the text field (user types in 25 but experiences a delay in seeing 25 in the input field for example). I think its still valuable to have a debounced handler instead of only updating the state.value variable on input blur, but I think we are still going to need the extra inputDisplayValue variables in the state because we want the value displayed in the text field to change instantaneously but delay the slider change slightly as the user types into the text field, what do you think?

prodonjs: Ahh I see I see. Then it probably makes sense to do what you suggest here. I'll leave it to you.

jupyterlab_gcedetails/src/components/input_slider.tsx
```diff
@@ -0,0 +1,94 @@
...
+    marginLeft: '10px',
+  },
+  label: {
+    display: 'block',
+    minWidth: '100px',
+  },
+});
+
+export function InputSlider(props: InputSliderProps) {
+  const { label, value, handleSliderChange, possibleValues } = props;
```
prodonjs, line 55: Just a nit, but is there any reason you don't unpack all of these props in one statement?

jupyterlab_gcedetails/src/data.ts
prodonjs: Yes, let's go with that.

jupyterlab_gcedetails/src/components/input_slider.tsx
yaminahq: The first set are for the slider and the second set are for the input so I was just making the distinction while i was working on this but I guess there is no good reason to do it separately, I'll change it :)

PR #90, opened 2020-07-06
Display comment threads

mkalil:
1) Modify component for displaying a comment to include "reply" and "expand thread" buttons. "Expand thread" toggles and renders additional comment components that are replies to the parent or top level comment. [Demo](https://drive.google.com/file/d/1G6xwp1pbfePx3uonDErBCKfDvCDk1RiE/view?usp=sharing) 
2) Add handler for fetching code review comments. This returns the original JSON object returned by "git appraise show -json" but modifies the comments value to only include comments corresponding to the file path passed into the handler.


jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -63,16 +63,35 @@ def get_comments_for_path(self, file_path_from_repo_root, git_root_dir):
...
+		"""
+		Returns the JSON for the current code review with comments for the requested file path.
+
+		Returns None of there are no review comments for the file path.
+		"""
+		#self.appraise_pull(current_path) #pull new comments from remote repo
+		review_string = self.run(git_root_dir, 'appraise', 'show',
+									'-json')
+		review_json = json.loads(review_string)
+		if "comments" in review_json.keys():
```
ojarjur, line 84: This condition doesn't seem right.

If there is a review with no comments, then you will return `None`, but if there is a review with no comments on this file, then you will return a review with an empty comments field.

I would suggest re-structuring this code somewhat:

1. Change the `else` clause to be a stand-alone condition that short-circuits the rest of the method;

```py
        if not review_json:
            return None
```

2. Then, reduce the indentation of the rest of the code.
3. Instead of explicitly checking if `comments` is in the dictionary, fetch the comments using `review_json.get("comments", [])`. That way the code is slightly simpler and you also treat the conditions of no comments vs no-matching-comments identically.

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -63,16 +63,35 @@ def get_comments_for_path(self, file_path_from_repo_root, git_root_dir):
...
+			comments_on_file = []
+			for item in review_json["comments"]:
+				location = item["comment"]["location"]
+				if "path" in location.keys() and location["path"] == file_path_from_repo_root:
```
ojarjur, line 88: This can be written as:

```py
if location.get("path", "") == file_path_from_repo_root:
```

PR #91, opened 2020-07-06
[Optimizer] add sidebar

kpfromer:
This adds the sidebar for the optimizer project that allows users to navigate to study details and main dashboard.

# Images

image


jupyterlab_caip_optimizer/src/components/sidebar_widget.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+          <Button color="primary" onClick={() => openDashboard()}>
+            Main Dashboard
+          </Button>
+        </Box>
+      </Box>
+
+      {!!error && <Typography color="error">{error}</Typography>}
+      {loading && <CircularProgress />}
+      {!!studies && studies.length > 0 && (
```
angelakuo, line 103: Maybe we should display text (e.g. "No studies found") if it's an empty list?

kpfromer: Here's what it looks like now: image

angelakuo: Not sure if we'd want to keep the error message in if there are correctly no studies loaded (feel free to address this in a separate PR)

PR #94, opened 2020-07-08
[BigQuery] Styling for details panel

CynthiaXJia:
Styling of details panel to resemble BigQuery web UI. Beyond using Material UI to make things look nicer, this PR includes the following changes:

- adds circular loading panel
- pulls shared details between tables and datasets into a separate component

SCREENSHOT BEFORE:
detailsbefore

SCREENSHOT AFTER:  
detailsafter


jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -75,7 +75,7 @@ def get_dataset_details(client, dataset_id):
   return {
       'details': {
           'id': "{}.{}".format(dataset.project, dataset.dataset_id),
-          'display_name': dataset.friendly_name,
+          'name': dataset.dataset_id,
```
CynthiaXJia, line 78: Following the BQ web UI, we display the `dataset_id` for datasets rather than `friendly_name`

jupyterlab_bigquery/src/components/details_panel/dataset_details_panel.tsx
```diff
@@ -24,17 +32,10 @@ export default class DatasetDetailsPanel extends React.Component<Props, State> {
       hasLoaded: false,
       isLoading: false,
       details: { details: {} } as DatasetDetails,
+      rows: [],
     };
   }
 
-  async componentDidMount() {
```
CynthiaXJia, line 30: We have a few of these empty `componentDidMount` functions scattered around, I've deleted this one and the one in `table_details_panel.tsx` since they aren't doing anything right now.

jupyterlab_bigquery/src/components/details_panel/dataset_details_panel.tsx
```diff
@@ -49,7 +50,25 @@ export default class DatasetDetailsPanel extends React.Component<Props, State> {
       const details = await this.props.datasetDetailsService.listDatasetDetails(
         this.props.dataset_id
       );
-      this.setState({ hasLoaded: true, details });
+
```
CynthiaXJia, line 53: Here and in `table_details_panel.tsx` I pull out the details that need to be shown in the striped table format and store them in `rows` to be passed to the common `details_panel.tsx`

PR #95, opened 2020-07-08
[BigQuery] Styling for query editor results

CynthiaXJia:
Adds ability to see first and last page of query results, as well as row numbers.  

resultsstyling


jupyterlab_bigquery/src/components/query_editor/query_editor_tab/query_editor_tab.tsx
```diff
@@ -15,8 +16,9 @@ class QueryEditorTab extends React.Component<QueryEditorTabProps, {}> {
 
   render() {
     return (
-      <div>
+      <div style={{ height: '100%' }}>
```
CynthiaXJia, line 19: Since right now `query_editor_tab.tsx` only acts as a wrapper for `QueryTextEditor` and `QueryResults`, I'm hoping to combine `query_editor_tab` into the overall `query_editor_tab_widget` after syncing with @MRzNone

cbwilkes:
It might be a good idea to circle back and separate the styles out into style classes in some places

jupyterlab_bigquery/src/components/query_editor/query_editor_tab/query_editor_tab.tsx
MRzNone: Decision being leave it as it is, for 2 reasons:
1, clean structure, since all widgets follows widget importing react stuff
2, Need to keep separate to use redux, since

PR #96, opened 2020-07-08
[Optimizer] add dashboard

kpfromer:
# Adds

This adds the main dashboard that is apart of feature 1.

# Photos

image
image


angelakuo:
What do you think about moving the "Create Study" button above the table so it's more visible?

jupyterlab_caip_optimizer/src/components/dashboard.tsx
```diff
@@ -0,0 +1,209 @@
...
+  // https://stackoverflow.com/questions/59648434/material-table-typeerror-cannot-add-property-tabledata-object-is-not-extensibl
+  const mappedStudies: MappedStudy[] = studies
+    ? studies.map(study => ({
+        name: study.name,
+        createTime: study.createTime,
+        state: study.state,
+        metrics: study.studyConfig.metrics
+          .map(
+            metric =>
+              `${capitalizeFirstLetter(metric.goal.toLowerCase())} "${
```
angelakuo, line 159: I would move the toLowerCase() into the helper function and rename it something like "makeReadable"

kpfromer:
Yeah I agree about moving the "Create Study" button. It looked out of place. Here's what it looks like now: 
image

PR #100, opened 2020-07-09
Automatic refresh

mkalil:
- Add configurable traitlet for the refresh interval
- Add handler to fetch the configured value
- Add handler to invoke 'git appraise pull' to fetch new comments from the configured remote
- Add setInterval() to invoke the remote fetch handler
- Fix getters for detached and review comments so that the UI will refresh even when the file clicked on has no comments

### DIFF
--- a/jupyterlab_comments/README.md
+++ b/jupyterlab_comments/README.md
@@ -55,5 +55,11 @@ jupyter labextension uninstall jupyterlab-comments
 Add the following lines to your Jupyter config file (e.g. jupyter_notebook_config.py):
 ```bash
 c.Git.remote = '${NAME_OF_REMOTE_REPOSITORY}'
+
+c.Refresh.interval = '${REFRESH_INTERVAL_SECONDS}'
 ```
-(The name of the remote repository defaults to 'origin' if no value is configured)
+1) The name of the remote repository defaults to 'origin' if no value is configured.
+
+2) New comments are pulled from the current repository at a constant interval which can be configured. The default interval is 10 seconds.
+
+
--- a/jupyterlab_comments/jupyterlab_comments/__init__.py
+++ b/jupyterlab_comments/jupyterlab_comments/__init__.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from notebook.utils import url_path_join
-from jupyterlab_comments.handlers import DetachedCommentsHandler, ReviewCommentsHandler, RefreshIntervalHandler
+from jupyterlab_comments.handlers import DetachedCommentsHandler, ReviewCommentsHandler, RefreshIntervalHandler, PullFromRemoteRepoHandler
 
 
 def load_jupyter_server_extension(nb_server_app):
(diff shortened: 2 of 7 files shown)


jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -38,8 +38,8 @@ def status(self, repo_path):
 		status = self.run(repo_path, 'status').decode('utf-8')
 		return status
 
-	def appraise_pull(self, repo_path):
-		self.run(repo_path, 'appraise', 'pull', self.remote)
+	def appraise_pull(self, git_root_dir):
+		self.run(git_root_dir, 'appraise', 'pull', self.remote)
```
ojarjur, line 42: We should probably implement some sort of queue to make sure that we don't have multiple `git appraise pull` commands happening in the same repository at the same time.

The `git appraise pull` command does not do any sort of locking to prevent two invocations happening at the same time (although, now that I think of it, it probably should).

It would be reasonable to push that change off to a subsequent PR, but I do suspect we will need to eventually do it.

jupyterlab_comments/README.md
```diff
@@ -55,5 +55,11 @@ jupyter labextension uninstall jupyterlab-comments
 Add the following lines to your Jupyter config file (e.g. jupyter_notebook_config.py):
 ```bash
 c.Git.remote = '${NAME_OF_REMOTE_REPOSITORY}'
```
gogasca, line 57: nit: Example: do NAME of Repo is the Repo URL? (Example: https://github.com/GoogleCloudPlatform/ai-platform-samples.git) ?

jupyterlab_comments/jupyterlab_comments/__init__.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from notebook.utils import url_path_join
-from jupyterlab_comments.handlers import DetachedCommentsHandler, ReviewCommentsHandler
+from jupyterlab_comments.handlers import DetachedCommentsHandler, ReviewCommentsHandler, RefreshIntervalHandler, PullFromRemoteRepoHandler
```
gogasca, line 16: nit: Import grouping is "recommended" but not enforced:

```
from jupyterlab_comments.handlers import DetachedCommentsHandler 
from jupyterlab_comments.handlers import ReviewCommentsHandler 
from jupyterlab_comments.handlers import RefreshIntervalHandler
from jupyterlab_comments.handlers import PullFromRemoteRepoHandler
```

jupyterlab_comments/jupyterlab_comments/handlers.py
```diff
@@ -87,3 +90,35 @@ def post(self):
...
         self.finish(
             'Add a detached comment for a specific file (unimplemented)')
+
+class RefreshIntervalHandler(APIHandler):
+
+    def get(self):
+        try:
+            interval = refresh.get_interval()
+            self.finish(json.dumps({"interval" : interval}))
+        except Exception as e:
```
gogasca, line 100: nit: Exception too broad? what exceptions does refresh.get_interval() or self.finish may generate?

jupyterlab_comments/README.md
mkalil: This is intended to be the short name of the remote repository to push to (if the user has more than one remote repository configured), such as 'origin' or 'upstream' -- like the output of the 'git remote' command. I'll change the README so it's more clear!

jupyterlab_comments/jupyterlab_comments/__init__.py
mkalil: This change will be incorporated in the next PR.

jupyterlab_comments/jupyterlab_comments/handlers.py
mkalil: This change will be incorporated in the next PR.

jupyterlab_comments/jupyterlab_comments/git_commands.py
mkalil: This change will be incorporated in a future PR.

PR #105, opened 2020-07-10
[Optimizer] add snackbar

kpfromer:
Adds a widget and redux action to open a helpful snackbar message. This can be used for errors and informing the user about actions in the app. Styles were [taken from material ui](https://material-ui.com/components/snackbars/#customized-snackbars). 

Severities:

- `success`
- `info`
- `warning`
- `error`

# Images

image
image
image
image


junhyeokkwak:
It seems like there are still merge conflicts?

jupyterlab_caip_optimizer/src/components/snackbar_widget.tsx
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const SnackbarUnwrapped = ({
+  open,
+  message,
+  severity,
+  close,
+}: Props) => {
+  const handleClose = (event?: React.SyntheticEvent, reason?: string) => {
+    if (reason === 'clickaway') {
```
junhyeokkwak, line 31: What does `'clickaway'` refer to? Does it mean user clicks on somewhere outside the snackbox?

jupyterlab_caip_optimizer/src/store/snackbar.ts
```diff
@@ -0,0 +1,53 @@
...
+  },
+});
+
+const wait = async (time: number) =>
+  new Promise(resolve => setTimeout(() => resolve(), time));
+
+export const createSnack = (
+  message: string,
+  severity: OpenAction['severity'] = 'info',
+  waitMs = 2000
```
junhyeokkwak, line 40: could we make `waitMs` an optional argument with default value so that we can change if we want to? I'd assume for longer error messages, 2 seconds might be a bit short. We can still have 2000ms as default.

kpfromer: Correct me if I am wrong, but `waitMs = 2000` should default to 2000ms.

jupyterlab_caip_optimizer/src/components/snackbar_widget.tsx
kpfromer: This was from the [snackbar demo on material ui](https://material-ui.com/components/snackbars/#customized-snackbars), but to my knowledge `clickaway` refers to when a user clicks off of the message to any other element on the page.

jupyterlab_caip_optimizer/src/store/snackbar.ts
kpfromer: https://material-ui.com/api/snackbar/ `onClose` property.

junhyeokkwak: oops sorry you're right. I didn't catch that it was inside the argument bracket.

jupyterlab_caip_optimizer/src/components/snackbar_widget.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import { Snackbar as MaterialSnackbar } from '@material-ui/core';
+import { Alert as MuiAlert, AlertProps } from '@material-ui/lab';
```
angelakuo, line 3: Assuming the Mui stands for MaterialUi, maybe we could rename it to MaterialAlert (similar to how MaterialSnackbar is renamed)?

jupyterlab_caip_optimizer/src/store/snackbar.ts
```diff
@@ -0,0 +1,53 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+import { AppThunk } from './store';
+
+export interface OpenAction {
+  severity: 'error' | 'warning' | 'info' | 'success';
```
angelakuo, line 5: Seems like something we can make into an Enum? Or a type

PR #106, opened 2020-07-13
[BigQuery] Table preview

CynthiaXJia:
Adds table preview under the `preview` tab in the table details window.  

tablepreview

### DIFF
--- a/jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
+++ b/jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
@@ -118,6 +118,14 @@ def get_table_details(client, table_id):
       }
   }
 
+def get_table_preview(client, table_id):
+  table = client.get_table(table_id)
+  rows = client.list_rows(table, max_results=100)
+  return {
+    'fields': [field.name for field in rows.schema],
+    'rows': [[value.__str__() if isinstance(value, datetime.date) else value for value in row.values()] for row in rows]
+  }
+
 class ListHandler(APIHandler):
   """Handles requests for Dummy List of Items."""
   bigquery_client = None
@@ -170,6 +178,27 @@ class TableDetailsHandler(APIHandler):
       post_body = self.get_json_body()
 
       self.finish(get_table_details(self.bigquery_client, post_body['tableId']))
+
+    except Exception as e:
+      app_log.exception(str(e))
+      self.set_status(500, str(e))
+      self.finish({
+          'error': {
+              'message': str(e)
+          }
+      })
+
+class TablePreviewHandler(APIHandler):
+  """"Handles request for table preview."""
+  bigquery_client = None
+
+  @gen.coroutine
+  def post(self, *args, **kwargs):
+    try:
+      self.bigquery_client = create_bigquery_client()
+      post_body = self.get_json_body()
+
+      self.finish(get_table_preview(self.bigquery_client, post_body['tableId']))
 
     except Exception as e:
       app_log.exception(str(e))
--- a/jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
+++ b/jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
@@ -6,7 +6,7 @@ import json
 import re
 import tornado.gen as gen
 import os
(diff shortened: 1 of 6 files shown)


jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -179,3 +187,24 @@ def post(self, *args, **kwargs):
               'message': str(e)
           }
       })
+
```
CynthiaXJia, line 190: Tests for this and other handlers will be coming up once @hwing777 is done reorganizing the handlers files

jupyterlab_bigquery/src/components/details_panel/table_details_panel.tsx
```diff
@@ -45,6 +46,14 @@ export default class TableDetailsPanel extends React.Component<Props, State> {
...
   }
 
+  componentDidMount() {
+    this.mounted = true;
+  }
+
+  componentWillUnmount() {
+    this.mounted = false;
+  }
+
```
CynthiaXJia, line 56: The `mounted` flag will check whether the tabs are mounted before updating state. This is to ensure we won't try to update an unmounted component and prevents the `Can't perform a React state update on an unmounted component` warning in the console.

jupyterlab_bigquery/src/components/details_panel/table_preview.tsx
```diff
@@ -0,0 +1,167 @@
...
+  TablePagination,
+} from '@material-ui/core';
+
+import LoadingPanel from '../loading_panel';
+import {
+  TableDetailsService,
+  TablePreview,
+} from './service/list_table_details';
+
+const localStyles = stylesheet({
```
CynthiaXJia, line 19: Since these are very similar to the table styles for the query editor results, I will pull out a shared table style in a future PR

jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -118,6 +118,14 @@ def get_table_details(client, table_id):
       }
   }
 
+def get_table_preview(client, table_id):
+  table = client.get_table(table_id)
+  rows = client.list_rows(table, max_results=100)
+  return {
+    'fields': [field.name for field in rows.schema],
+    'rows': [[value.__str__() if isinstance(value, datetime.date) else value for value in row.values()] for row in rows]
```
CynthiaXJia, line 126: For now, this handles date objects along with strings, but I'm not sure what other objects might come through from BQ table entries that I need to handle.

CynthiaXJia: update: I will make changes to handle all possible schema types and add it as a commit to this pr

jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -89,13 +89,25 @@ def get_dataset_details(client, dataset_id):
...
-    } for field in schema]
+      })
+    if field.field_type == 'RECORD':
+      for record_entry in field.fields:
+        formatted_schema.append({
+          'name': "{}.{}".format(field.name, record_entry.name),
+          'type': record_entry.field_type,
+          'description': record_entry.description,
+          'mode': record_entry.mode
+        })
```
CynthiaXJia, line 108: Handles records, but only singly-nested records for now. A future PR will address deeply nested records.

jupyterlab_bigquery/jupyterlab_bigquery/handlers.py
```diff
@@ -114,10 +126,57 @@ def get_table_details(client, table_id):
           'link': table.self_link,
           'num_rows': table.num_rows,
           'num_bytes': table.num_bytes,
-          'schema': get_schema(table.schema)
+          'schema': format_detail_schema(table.schema)
       }
   }
 
+def format_preview_fields(schema):
```
CynthiaXJia, line 133: general question: is it better to format things like dates, bytes, etc here in the backend, or is it better to pass them to the frontend as a json string and then parse and format that in the frontend?

PR #107, opened 2020-07-13
[Optimizer] Add 'Create Study' page

junhyeokkwak:
Added a 'Create Study' page that allows users to create an Optimizer study by filling the input fields. This page is activated by clicking on the 'Create Study' button on the main dashboard and a successful creation of study changes the view to 'Study Details' page.

## Features
* Only the type-specific value input fields get activated for different parameter/metric types
* Buttons get activated only when the required fields are filled
* Redux store and sidebar gets updated after successful creation of study


jupyterlab_caip_optimizer/src/components/create_study.tsx
```diff
@@ -0,0 +1,603 @@
...
+  createStudyAndLoad: (study: Types.Study) => void;
+}
+
+const mapDispatchToProps = dispatch => ({
+  createStudyAndLoad: (study: Types.Study) =>
+    dispatch(createStudy(study)).then((study: Types.Study) =>
+      dispatch(
+        setView({
+          view: 'studyDetails',
+          studyId: study.name, // might need to prettify?
```
kpfromer, line 43: This should work without needing to prettify.

jupyterlab_caip_optimizer/src/components/create_study.tsx
```diff
@@ -0,0 +1,603 @@
...
+  const [paramType, setParamType] = React.useState<Types.ParameterType>(
+    'PARAMETER_TYPE_UNSPECIFIED'
+  );
+  const paramTypes: DropdownItem[] = createDropdown(Types.ParameterTypeList);
+  const [paramMinVal, setParamMinVal] = React.useState('');
+  const [paramMaxVal, setParamMaxVal] = React.useState('');
+  const [paramValList, setParamValList] = React.useState([]);
+  const [paramValListString, setParamValListString] = React.useState('');
+  const [paramIsNew, setParamIsNew] = React.useState(true);
+  const [paramChipData, setParamChipData] = React.useState<ParameterChip[]>([]); // TODO: change this to empty list. for demo purposes
```
kpfromer, line 96: Is this TODO finished?

jupyterlab_caip_optimizer/src/components/sidebar.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
```
kpfromer, line 1: I think this was the temporary sidebar for testing. I would delete this file.

package.json
```diff
@@ -41,5 +41,9 @@
       "pre-commit": "npm run lint",
       "pre-push": "npm run lint && npm test"
     }
+  },
+  "dependencies": {
+    "@material-ui/core": "^4.10.1",
```
kpfromer, line 46: I don't think these need to be added here.

jupyterlab_gcedetails/src/details_widget.tsx
```diff
@@ -76,11 +76,7 @@ export class VmDetails extends React.Component<Props, State> {
     return (
       <span className={STYLES.container}>
         <span
-          className={classes(
-            STYLES.icon,
-            ICON_CLASS,
-            STYLES.interactiveHover
-          )}
+          className={classes(STYLES.icon, ICON_CLASS, STYLES.interactiveHover)}
```
kpfromer, line 79: I might remove these formatting changes for the other libraries. It's strange that the linter did not fix it for those who edited these files.

jupyterlab_caip_optimizer/src/components/create_study.tsx
junhyeokkwak: yeah I'll edit this

jupyterlab_caip_optimizer/src/components/sidebar.tsx
junhyeokkwak: done

package.json
junhyeokkwak: done

jupyterlab_caip_optimizer/src/types.ts
```diff
@@ -6,41 +6,54 @@ export interface AsyncState<T> {
...
-}
+/**
+ * 'Const assignment" used to control the string inputs in dropdown input fields.
+ * String enums don't support reverse mapping, and plain string literal types aren't iterable.
+ */
+export const GoalTypeList = [
+  'MAXIMIZE',
+  'MINIMIZE',
+  'GOAL_TYPE_UNSPECIFIED',
+] as const;
```
angelakuo, line 22: Is the `as const` necessary? That's surprising to me

junhyeokkwak: If there isn't `as const` here, the `typeof` in the next line will widen the types to `string`. `As const` makes sure that the literal types don't get widened so that when we try to extract types from the array, the type will be `'MAXIMIZE' | 'MINIMIZE' | 'GOAL_TYPE_UNSPECIFIED'` instead of `string`.
Reference: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

PR #108, opened 2020-07-13
[HWConfiguration] Lint fixes

prodonjs:


prodonjs:
> Hmm weird that these didn't get picked up on initial push

Yeah I think it depends on whether or not you run `npm run lint` before your commit or not. It runs in the Presubmit check but obviously that won't modify your code. Adding a pre-commit command that runs lint is probably a good idea.

prodonjs:
@angelakuo Can you approve this? I forgot that only FTEs can give Merge approval.

PR #112, opened 2020-07-14
[BigQuery] Showing Query Editor among jupyter cells and add support for multiple query editors

MRzNone:


MRzNone:
UI

Pipfile
```diff
@@ -9,6 +9,8 @@ python_version = "3.7"
 [packages]
 jupyterlab = "==1.2.15"
 pandas = "*"
+ipywidgets = "*"
+traitlets = "*"
```
MRzNone, line 13: Need for ipywidget UI definition

jupyterlab_bigquery/jupyterlab_bigquery/query_incell_editor.py
```diff
@@ -0,0 +1,16 @@
+from ipywidgets import DOMWidget
+from traitlets import Unicode
+
+# TODO: refactor name and version to sync with front 
+module_name = 'bigQuery-query-incell-editor'
```
cbwilkes, line 5: Let's not mix the camelCase and kebal-case. snake_case would be preferred. Can this be rename to `bigquery_query_incell_editor`?

Same thing for the other locations.

README.md
```diff
@@ -50,6 +50,8 @@ environment.
    JupyterLab environment in development mode. Afterwards, `cd` back to the root
    of the repository.
 
+   - Please also run `jupyter labextension install @jupyter-widgets/jupyterlab-manager` if your extension uses ipywidget. Currently, big_query extension is using it.
```
cbwilkes, line 53: Can we accomplish this a better way? 

Can we just add a dependency `"@jupyter-widgets/jupyterlab-manager": "1.1",` to https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/master/jupyterlab_bigquery/package.json

MRzNone: I tried installing that from `npm install @jupyter-widgets/jupyterlab-manager@1.1`.  It did not work.  I think `jupyter labextension install` is doing some extra things to setup.  Did not find resources `jupyter labextension install` vs `npm install`.

jupyterlab_bigquery/jupyterlab_bigquery/query_incell_editor.py
MRzNone: I fixed the module_name.  The other names in `Unicode` need to correspond to class names in front-end, so I did not change them.  Should I?

README.md
MRzNone: I can make this a sub-command in `jupyterlab_bigquery/package.json`.  However, I am personally against it for 2 reasons:
1) jupyter lab extension manager is kind not working so well.  Different jupyter versions needs different version of extension manager.  We can mess user's local extension manager if we install it here
2) other packages (ipywidgets) do not have the installed automatically.

jupyterlab_bigquery/src/components/query_editor/query_editor_incell/query_editor_incell_widget.tsx
```diff
@@ -0,0 +1,48 @@
...
+  static serializers: ISerializers = {
+    ...DOMWidgetModel.serializers,
+    // Add any extra serializers here
+  };
+
+  static model_name = 'QueryIncellEditorModel';
+  static model_module = 'bigQuery-query-incell-editor';
+  static model_module_version = '0.0.1';
+  static view_name = 'QueryIncellEditorView'; // Set to null if no view
+  static view_module = 'bigQuery-query-incell-editor'; // Set to null if no view
```
cbwilkes, line 37: Can bigQuery-query-incell-editor be renamed to bigquery_query_incell_editor here too?

README.md
cbwilkes: Ok I we will just leave it here, even though I wish there was a cleaner way.

jupyterlab_bigquery/jupyterlab_bigquery/query_incell_editor.py
cbwilkes: It would be nice if all the names are consistent. Preferably change `bigQuery-query-incell-editor` to `bigquery_query_incell_editor` everywhere you can.

PR #114, opened 2020-07-14
[HardwareConfiguration] Adding machine type drop down select component

yaminahq:
Adding NestedSelect component which allows a user to select a machine type.

Screenshot 2020-07-13 at 11 40 26 PM

### DIFF
--- a/jupyterlab_gcedetails/src/components/machine_type_select.tsx
+++ b/jupyterlab_gcedetails/src/components/machine_type_select.tsx
@@ -94,19 +94,19 @@ class HeaderItem extends React.Component<HeaderItemProps, HeaderItemState> {
   }
 }
 
-/* Menu body component */
+/* Nested select body component */
 
-interface MenuBodyProps {
+interface NestedSelectBodyProps {
   onSelect: (o: Option) => void;
   selectedOption: Option;
 }
 
-interface MenuBodyState {
+interface NestedSelectBodyState {
   displayedOptions: Option[];
 }
 
-class MenuBody extends React.Component<MenuBodyProps, MenuBodyState> {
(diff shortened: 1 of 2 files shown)


jupyterlab_gcedetails/src/components/machine_type_select.tsx
```diff
@@ -0,0 +1,257 @@
...
+    );
+  }
+
+  render() {
+    return (
+      <Paper elevation={5} square>
+        <div className={STYLES.menuContainer}>
+          <Grid container spacing={0} alignItems="center">
+            <Grid item xs>
+              {this.headersList}
```
prodonjs, line 177: Is this meant to reference the method and not actually call it?

jupyterlab_gcedetails/src/components/machine_type_select.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+class NestedSelectBody extends React.Component<
+  NestedSelectBodyProps,
+  NestedSelectBodyState
+> {
+  constructor(props: NestedSelectBodyProps) {
+    super(props);
+
+    this.state = {
+      displayedOptions: MACHINE_TYPES[0].configurations,
```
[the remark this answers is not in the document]
yaminahq: that's a good point, I was actually implementing this component with the fact that it may be valuable as a shared component in mind so the change should be easy enough - I'll add it in!

jupyterlab_gcedetails/src/components/machine_type_select.tsx
yaminahq: good catch!

PR #118, opened 2020-07-16
[HardwareConfiguration] Modifying hardware scaling form to use drop down …

yaminahq:
…machine type select and updated select input

[Video Demo](https://drive.google.com/file/d/1NbCjiwrYAwXNuA-AtIPcR3VnFzRJMNjc/view?usp=sharing)

image

This PR:
- Modifies the Hardware Scaling form to use the dropdown machine type select
- Adds SelectInput component that matches the MaterialUI styling and updates the Hardware Scaling Form to  use these select inputs
- Deletes the code corresponding to the Sliders. This includes code in data.ts, input_slider.tsx, synchronized_slider.tsx


jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -114,16 +106,11 @@ export class HardwareScalingForm extends React.Component<Props, State> {
...
   }
 
-  private onBaseMachineTypeChange(event: React.ChangeEvent<HTMLSelectElement>) {
-    this.setState({
-      baseMachineType: event.target.value,
-      cpuValue: this.getMachineType(event.target.value).configurations[0].cpu,
-      memoryValue: this.getMachineType(event.target.value).configurations[0]
-        .memory,
-    });
+  private onMachineTypeChange(machineType: Option) {
```
prodonjs, line 109: Could just make this an inline handler now that the rest of the logic has been removed.

jupyterlab_gcedetails/src/data.ts
```diff
@@ -391,3 +337,8 @@ export const STYLES = stylesheet({
     boxShadow: 'inset 0 -1px 0 0 var(--jp-border-color0)',
   },
 });
+
+export const TEXT_STYLE = {
+  fontFamily: BASE_FONT.fontFamily as string, //'var(--jp-ui-font-family)'
```
prodonjs, line 342: Are the explicit type annotations necessary here?

jupyterlab_gcedetails/src/components/select_input.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return <ArrowDropDownIcon className={STYLES.icon} />;
+};
+
+export function SelectInput(props: Props) {
+  const { label, name, value, options, onChange } = props;
+
+  return (
+    <div>
+      {label && <label>{label}</label>}
+      <TextField
```
prodonjs, line 53: Just curious, what's the advantage of the TextField vs. the Select React material component?

yaminahq: I'm specifically using it to get the styling of the machine type nested select component that I built and this select input to match better. I originally tried using the Select React material component but some of their default styling made this harder to do. It may also make sense to extract the textfield out into a wrapper component and have the machine type nested select component and the select input component use that instead so styling across all future inputs for this extension are consistent but I think for now this is fine, what do you think?

jupyterlab_gcedetails/src/data.ts
yaminahq: Because I'm passing this const to Material ui components directly as props, it starts complaining if the type is not explicitly set to string (as `BASE_FONT.fontFamily` has type `string | string[]`). I can probably figure out a workaround though if this is a concern? :)

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
yaminahq: Good point!

jupyterlab_gcedetails/src/components/select_input.tsx
prodonjs: SGTM

jupyterlab_gcedetails/src/data.ts
prodonjs: Nah this good.

PR #119, opened 2020-07-16
[BigQuery] Make entire details panel scrollable

CynthiaXJia:
In a [previous PR](https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/99) I made just the schema scrollable, but since it is currently on the same panel as the details, the entire panel should be scrollable to account for smaller screens.

### DIFF
--- a/jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
+++ b/jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
@@ -106,6 +106,8 @@ function TabPanel(props: TabPanelProps) {
         minHeight: 0,
         flexDirection: 'column',
         display: value !== index ? 'none' : 'flex',
+        overflowX: value === TabInds.preview ? 'auto' : 'hidden',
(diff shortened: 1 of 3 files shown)


jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
```diff
@@ -106,6 +106,8 @@ function TabPanel(props: TabPanelProps) {
         minHeight: 0,
         flexDirection: 'column',
         display: value !== index ? 'none' : 'flex',
+        overflowX: value === TabInds.preview ? 'auto' : 'hidden',
+        overflowY: 'auto',
```
CynthiaXJia, line 110: The reason for this weird check is that mui grids have a spacing property that sets the padding rather than margin, which causes the content to overflow slightly and leads to an annoying horizontal scroll bar. In the details panel you'll never really need an x scrollbar but in the preview you will likely have both x and y so that's why you see that here.

PR #121, opened 2020-07-16
[uCAIP] Import data from pandas dataframe feature

josie-garza:
- relies on pr #102 
- converts the pandas dataframe to a csv and then sends the bytes to be uploaded as a dataset
- uses the app Service Manager to get the running sessions and ClientSession to connect to the selected kernel
- does not shutdown the kernel but makes it busy while the process occurs
- includes loading and error snackbars in the bottom right

Main files to look at: import_data.tsx, kernel_model.ts

image
image


jupyterlab_automl/src/service/kernel_model.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!this._session || !this._session.kernel) {
+      return;
+    }
+    const path = './' + df + '.csv';
+    const run =
+      "import base64;df.to_csv('" +
+      path +
+      "', index=False);base64.b64encode(open('" +
+      path +
+      "', 'r').read().encode('utf8')).decode('utf-8')";
```
josie-garza, line 77: - this goes into the kernel and turns the dataframe to a csv and converts that to bytes to create a dataset
- the name of the csv file is the name of the variable the user put in the import data form
- should this file be deleted from the local file system once the bytes have been uploaded?

PR #125, opened 2020-07-16
[Optimizer] StudyDetails page implemented

junhyeokkwak:
## Overview
- Implementation of "StudyDetails" page
- Link "StudyDetails" with other views
- Added 'suggestTrials' and 'visualizeTrials' views and corresponding actions/reducers to redux store
- Moved util functions to a separate file
## Notes
- Trial table to be added after the "Suggest Trials" page implementation for design consistency


jupyterlab_caip_optimizer/src/store/view.ts
```diff
@@ -19,9 +19,13 @@ export const viewSlice = createSlice({
   } as ViewState,
   reducers: {
     setView: (state, action: PayloadAction<ViewType>) => {
-      if (action.payload.view === 'studyDetails') {
+      if (
```
kpfromer, line 22: I cleaned this up on my branch. It should make this reducer less messy.

jupyterlab_caip_optimizer/src/components/study_details.tsx
```diff
@@ -0,0 +1,297 @@
...
+                      <StyledTableRow key={`param-row-${index}`}>
+                        <StyledTableCell
+                          key={`param-row-${index}-cell-name`}
+                          component="th"
+                          scope="row"
+                        >
+                          {row.name}
+                        </StyledTableCell>
+                        <StyledTableCell
+                          key={`param-row-${index}-cell-type`}
```
kpfromer, line 250: You only need a key for the top most element in a map function. These sub `key` values are not needed.

jupyterlab_caip_optimizer/src/store/view.ts
junhyeokkwak: Then I guess I'll leave this as it is for now, and can be cleaned in your PR?

jupyterlab_caip_optimizer/src/components/study_details.tsx
junhyeokkwak: fixed!

jupyterlab_caip_optimizer/src/components/study_details.tsx
```diff
@@ -0,0 +1,292 @@
...
+
+interface FormattedParam {
+  name: string;
+  type: Types.ParameterType;
+  values?: number[] | string[];
+  minValue?: number | string;
+  maxValue?: number | string;
+}
+
+interface FormattedStudy {
```
angelakuo, line 55: nit; I think I would just grab the actual types or declare as strings, this probably makes sense if the types will be changed somewhere down the line but that generally shouldn't happen with public APIs

jupyterlab_caip_optimizer/src/components/study_details.tsx
```diff
@@ -0,0 +1,292 @@
...
+  name: Types.Study['name'];
+  objective: string;
+  algorithm: Types.StudyConfig['algorithm'];
+  state: Types.Study['state'];
+  createTime: Types.Study['createTime'];
+  parameters: FormattedParam[];
+}
+
+function formatParams(params: Types.ParameterSpec[]): FormattedParam[] {
+  const formattedParams: FormattedParam[] = params.map(param => {
```
angelakuo, line 65: nit; You can probably just put the return here instead of declaring a const beforehand

jupyterlab_caip_optimizer/src/components/study_details.tsx
junhyeokkwak: I'm a bit confused about what you're suggesting regarding grabbing the actual types. Right now, I'm grabbing all the actual types except the objective which is a new param that isn't in `Types.Study` type. Would you prefer all the params to be plain strings instead?

junhyeokkwak: Yeah I can do that, but wouldn't this way be better since if we have to change Types.Study at some point due to API change, the types would automatically align?

jupyterlab_caip_optimizer/src/components/study_details.tsx
junhyeokkwak: Fixed it :)

PR #126, opened 2020-07-17
[BigQuery] IPython magic for incell editor

MRzNone:
IPython magic for incell editor

- load query from cell into query editor
- bug fix: error for empty query
- query editor check SQL query on load
- communication between python and TS


MRzNone:
image

PR #127, opened 2020-07-17
[BigQuery] PagedAPI testing and clear generators 

MRzNone:


MRzNone:
refactorrf `id` to `job_id` since `id` is python reserved function

jupyterlab_bigquery/jupyterlab_bigquery/tests/pagedAPI_handler/pagedAPI_handler_test.py
```diff
@@ -0,0 +1,131 @@
...
+PATH = '/gcp/v1/bigquery'
+
+
+class DummyPageAPI(PagedAPIHandler):
+
+  def query(self, request_body, page_size):
+    return request_body, page_size
+
+
+class TestPagedAPI_handler(unittest.TestCase):
```
cbwilkes, line 19: You should split this up differently, so that you can tell when one case fails.

Also add a unit test for clear_generators

Something like:

```
class TestPagedAPIMultplex_handler(unittest.TestCase):
  def setUp(self, fake_super):
    self. clear_mocks()

  def test_post_onstart():
    ...
  def test_post_oncontinue():
    ...
  def test_post_oncancel():
    ...
  def test_post_empty():
    ...

class TestPagedAPIQuery_handler(unittest.TestCase):

  def setUp(self, fake_super):
    self. reset_mocks()

  def test_normal():
    ...

  def test_error():
    ...

class TestClearGenerator(unittest.TestCase):
  def test_clear_generators(self):
    ..
```

PR #133, opened 2020-07-18
[BigQuery] Create search feature

hwing777:
Creates the search feature using the Data Catalog API


cbwilkes:
Seems like there are a lot of conflicts. Can these be resolved?

jupyterlab_bigquery/jupyterlab_bigquery/list_items_handler/create_handler.py
```diff
@@ -17,6 +17,22 @@ def _handle_exception(self, e):
     self.finish({"error": {"message": str(e)}})
 
 
+class Handlers():
+  """Singleton for handlers dictionary"""
+
+  _instance = {}
```
cbwilkes, line 23: You should set this to None, and create another variable dict to store the handlers.

Technically objects are dicts, but we should separate that out into a variable

jupyterlab_bigquery/src/components/list_items_panel/list_search_results.tsx
```diff
@@ -0,0 +1,209 @@
...
+          label: item.label,
+          onClick: () => item.handler(model),
+        }))}
+      >
+        <TreeItem
+          nodeId={model.id}
+          label={
+            <div>
+              <Typography>{model.name}</Typography>
+              <Typography style={{ fontSize: 12, color: 'gray' }}>
```
cbwilkes, line 138: Since these are duplicated can these be added to the localStyles?

PR #134, opened 2020-07-20
[uCAIP] View and list training pipelines

shunr:
Dependent on https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/102

A relatively complete changeset that adds the ability to list training pipelines and view pipeline details in a main area widget.
Refactors training pipelines and model representations

### DIFF
--- a/jupyterlab_automl/jupyterlab_automl/__init__.py
+++ b/jupyterlab_automl/jupyterlab_automl/__init__.py
@@ -3,6 +3,7 @@
 from notebook.utils import url_path_join
 from jupyterlab_automl.handlers import handlers
 from jupyterlab_automl.version import VERSION
+from jupyterlab_automl.public import *
 
 __version__ = VERSION
 
--- a/jupyterlab_automl/jupyterlab_automl/public.py
+++ b/jupyterlab_automl/jupyterlab_automl/public.py
@@ -1,0 +1,13 @@
+from jupyterlab_automl.service import AutoMLService
+
+
+def list_datasets():
+  return AutoMLService.get().get_datasets()
+
+
+def list_models():
+  return AutoMLService.get().get_models()
+
+
+def create_dataset_from_dataframe(display_name, df):
+  return AutoMLService.get().create_dataset_from_dataframe(display_name, df)
--- a/jupyterlab_automl/jupyterlab_automl/service.py
+++ b/jupyterlab_automl/jupyterlab_automl/service.py
@@ -251,18 +251,29 @@ class AutoMLService:
                                                   dataset=ds).result()
     return created
 
-  def create_dataset_from_file(self, display_name, file_name, file_data):
+  def _get_gcs_bucket(self):
     try:
-      bucket = self._gcs_client.create_bucket(self._gcs_bucket)
+      return self._gcs_client.create_bucket(self._gcs_bucket)
     except exceptions.Conflict:
       # Bucket already exists
-      bucket = self._gcs_client.bucket(self._gcs_bucket)
+      return self._gcs_client.bucket(self._gcs_bucket)
 
+  def create_dataset_from_file(self, display_name, file_name, file_data):
+    bucket = self._get_gcs_bucket()
     key = "{}-{}".format(str(uuid.uuid4()), file_name)
     # Possible improvement: prepend md5 hash of file instead of uuid to
     # skip uploading the same existing file
     decoded = base64.decodebytes(file_data.encode("ascii"))
     bucket.blob(key).upload_from_string(decoded)
+    return self.create_dataset(display_name,
+                               gcs_uri="gs://{}/{}".format(
+                                   self._gcs_bucket, key))
+
+  def create_dataset_from_dataframe(self, display_name, df):
+    bucket = self._get_gcs_bucket()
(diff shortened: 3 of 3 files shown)


jupyterlab_automl/src/components/list_resources_panel.tsx
```diff
@@ -417,32 +526,37 @@ export class ListResourcesPanel extends React.Component<Props, State> {
...
-    }
+    this.setState({ isLoading: true });
+    await Promise.all([
+      this.getDatasets(),
+      this.getModels(),
+      this.getPipelines(),
+    ]);
+    this.setState({ hasLoaded: true });
+    this.setState({ isLoading: false });
   }
```
josie-garza, line 537: should the refresh just update the resource type that the user is on (i.e. datasets, models, or training pipelines) to save time?

jupyterlab_automl/setup.py
```diff
@@ -55,7 +55,7 @@
     install_requires=[
         "google-cloud-storage>=1.24.1", "jupyterlab~=1.2.0",
         "gcp_jupyterlab_shared",
-        "pip @ https://storage.googleapis.com/google-cloud-aiplatform/libraries/python/0.1.0/google-cloud-aiplatform-0.1.0.tar.gz",
+        "google-cloud-aiplatform @ https://storage.googleapis.com/google-cloud-aiplatform/libraries/python/0.1.0/google-cloud-aiplatform-0.1.0.tar.gz",
```
prodonjs, line 58: I know Josie was asking me about this before I left on vacation, but were you ever able to confirm that this works properly when you do a fresh installation of the package on an actual CAIP Notebook?

jupyterlab_automl/src/components/list_resources_panel.tsx
```diff
@@ -121,6 +133,153 @@ export class ListResourcesPanel extends React.Component<Props, State> {
     }
   }
 
+  private getTableForResourceType(type: ResourceType) {
+    const sharedProps = {
+      isLoading: this.state.isLoading,
+      height: this.props.height - 80,
+      width: this.props.width,
+    };
+    switch (type) {
```
prodonjs, line 142: This is a pretty big switch statement. It looks like there might be some redundancy into the ListResourcesTable components that are returned from each block. Any chance the inner sections could be factored into other private methods so this method doesn't end up spanning so long? I think one of the rules of thumb that some of the Google style guides suggest is that if the method requires scrolling, it's a good candidate to split apart.

jupyterlab_automl/src/service/model.ts
```diff
@@ -53,12 +66,35 @@ export interface ModelEvaluation {
...
+          m: () => 'min',
+          s: () => 'sec',
+        },
+      },
+    });
+    data.elapsedTime = humanizer(Math.round(data.elapsedTime as number) * 1000);
+  }
+}
+
 export abstract class ModelService {
```
prodonjs, line 93: Just out of curiosity, what's the intent of the this being an abstract class with only static methods? Could you accomplish the same thing with a module containing only functions?

jupyterlab_automl/src/components/list_resources_panel.tsx
```diff
@@ -417,32 +526,37 @@ export class ListResourcesPanel extends React.Component<Props, State> {
...
-      this.setState({ isLoading: true });
-      await Promise.all([this.getDatasets(), this.getModels()]);
-      this.setState({ hasLoaded: true });
-    } catch (err) {
-      console.warn('Error retrieving datasets', err);
-    } finally {
-      this.setState({ isLoading: false });
-    }
+    this.setState({ isLoading: true });
+    await Promise.all([
```
prodonjs, line 530: Since you swallow any errors in the get* methods, there's no great way for a user to tell if something didn't refresh properly. Not a big issue for an MVP, but you may want to add a TODO to give some consideration to what can be surfaced if data doesn't get retrieved successfully.

jupyterlab_automl/setup.py
josie-garza: Yes. I checked earlier and it does work on a fresh installation on a CAIP Notebook.

jupyterlab_automl/src/components/model_properties.tsx
```diff
@@ -1,221 +1,35 @@
...
-            this.createData(
-              properties[i]['label'],
-              pipeline[properties[i]['name']]
-            )
-          );
-        }
-      }
-    }
-    return modelDetails;
+    this.getModelDetails(this.props.pipeline);
```
josie-garza, line 32: should this be `this.getPipeline()` because `this.getModelDetails()` does not exist in this scope?

jupyterlab_automl/src/components/model_properties.tsx
```diff
@@ -1,221 +1,35 @@
...
 }
 
 interface State {
   hasLoaded: boolean;
   isLoading: boolean;
-  modelDetails: any[];
-  transformationOptions: any[];
+  pipeline: Pipeline;
   open: boolean;
 }
```
josie-garza, line 18: I am a little confused why pipeline is a prop and also a state variable. We can probably get rid of both the `pipeline` and `open` state variables

PR #137, opened 2020-07-20
[uCAIP] Import from pandas dataframe

josie-garza:
- converts the pandas dataframe to a csv and then sends the bytes to be uploaded as a dataset
- uses the app Service Manager to get the running sessions and ClientSession to connect to the selected kernel
- does not shutdown the kernel but makes it busy while the process occurs
- includes loading and error snackbars in the bottom right

image
image


josie-garza:
- the name of the csv file is the name of the variable the user put in the import data form
- should the file be deleted from the local file system once the bytes have been uploaded?

PR #141, opened 2020-07-22
[BigQuery] Shared table and header styles

CynthiaXJia:
Creates a new shared `Header` and `BQTable` component that can be used across the extension.

### DIFF
--- a/jupyterlab_bigquery/src/components/query_editor/query_editor_tab/query_editor_results.tsx
+++ b/jupyterlab_bigquery/src/components/query_editor/query_editor_tab/query_editor_results.tsx
@@ -1,53 +1,24 @@
 // @ts-nocheck
 import React, { Component } from 'react';
 import { stylesheet } from 'typestyle';
-import {
-  Table,
-  TableHead,
-  TableRow,
-  TableCell,
-  TableBody,
-  TablePagination,
-  IconButton,
-} from '@material-ui/core';
-import {
-  KeyboardArrowLeft,
-  KeyboardArrowRight,
-  FirstPage,
-  LastPage,
-} from '@material-ui/icons';
 import { connect } from 'react-redux';
 import { QueryResult } from '../query_text_editor/query_text_editor';
 import { QueryId } from '../../../reducers/queryEditorTabSlice';
+import { Header } from '../../shared/header';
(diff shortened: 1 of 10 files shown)


jupyterlab_bigquery/src/components/query_editor/query_editor_tab/query_editor_results.tsx
```diff
@@ -59,67 +30,7 @@ interface QueryResultsState {
...
-      </IconButton>
-      <IconButton
-        onClick={handleLastPageButtonClick}
-        disabled={page >= Math.ceil(count / rowsPerPage) - 1}
-      >
-        <LastPage />
-      </IconButton>
-    </div>
-  );
+  location?: string;
```
CynthiaXJia, line 33: This prop specifies whether the editor is in the full window or in a notebook cell.

cbwilkes: Can you turn this into a type? 

Similar to:

```
type FULL_WINDOW = 'FULL_WINDOW';
type IN_CELL = 'IN_CELL'
type QueryEditorType = FULL_WINDOW | IN_CELL;

interface QueryResultsProps {
  queryResult: QueryResult;	  queryResult: QueryResult;
  queryId: QueryId;	  queryId: QueryId;
  editorType?: QueryEditorType;
}
```

Then other locations you can use:

```
if (this.props.editorType === FULL_WINDOW) { ... }
```

CynthiaXJia: It seems `this.props.editorType === FULL_WINDOW` won't work because `'IN_CELL' only refers to a type, but is being used as a value here.` Did you mean `this.props.editorType === 'FULL_WINDOW'`?

Is the idea to export the `FULL_WINDOW` and `IN_CELL` types for use elsewhere, or is it to more strongly enforce that we can't pass in any other strings besides these two?

cbwilkes: Sometimes it is better to refer to const rather than using string. But we can move forward with the strings. That should protect from setting the value to something other than a valid string.

PR #144, opened 2020-07-22
[BigQuery] paged query handler test

MRzNone:


jupyterlab_bigquery/jupyterlab_bigquery/tests/pagedAPI_handler/pagedQuery_handler_test.py
```diff
@@ -0,0 +1,247 @@
...
+
+    fake_client_query.side_effect = err
+    fake_job_config.return_value = dry_run_job_config_mock
+
+    gen = self.dummy_query_handler.query(request_body, page_size)
+
+    try:
+      next(gen)
+      self.fail()
+    except Exception as inst:
```
cbwilkes, line 140: Use `assertRaises` or `assertRaisesRegex` here and below instead:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaisesRegex

jupyterlab_bigquery/jupyterlab_bigquery/tests/pagedAPI_handler/pagedQuery_handler_test.py
```diff
@@ -0,0 +1,247 @@
...
+    fake_job_config.return_value = dry_run_job_config_mock
+
+    gen = self.dummy_query_handler.query(request_body, page_size)
+
+    actual_dry_run_job, actual_job_id = next(gen)
+
+    fake_job_config.assert_called_with(dry_run=True, use_query_cache=False)
+    fake_client_query.assert_called_with(query,
+                                         job_config=dry_run_job_config_mock)
+    self.assertEqual(actual_dry_run_job, dry_run_job)
```
cbwilkes, line 65: I generally prefer not to compare to a Mock object unless the data types are from an external library that is difficult to reproduce.

Is it possible to compare actual_dry_run_job to a real object instead of a mock. Or just assert that actual_dry_run_job.job_id == dry_run_job.job_id.

PR #145, opened 2020-07-22
Add comments handler

mkalil:
- Handler for adding new detached comments and pushing them to the remote repo
- Additional components: TextField for starting a new comment thread; expandable TextField for replying to a specific comment

### DIFF
--- a/jupyterlab_comments/jupyterlab_comments/git_commands.py
+++ b/jupyterlab_comments/jupyterlab_comments/git_commands.py
@@ -41,6 +41,9 @@ class Git(Configurable):
 	def appraise_pull(self, git_root_dir):
 		self.run(git_root_dir, 'appraise', 'pull', self.remote)
 
+	def appraise_push(self, git_root_dir):
+		self.run(git_root_dir, 'appraise', 'push')
+
 	def inside_git_repo(self, path_to_file):
 		"""
 		Return true if the path is inside a git repository.
@@ -93,8 +96,8 @@ class Git(Configurable):
 		return review_json
 
 
-	def add_comment(self, file_path, server_root):
-		pass
+	def add_detached_comment(self, file_path_from_repo_root, git_root_dir, comment_string):
+		self.run(git_root_dir, 'appraise', 'comment', '-d', '-m', comment_string, '-f', file_path_from_repo_root)
 
 	def get_previous_names(self, file_path, server_root):
 		# names_string = run('log', '--follow', '--name-only', '--pretty=format:""', file_path)
--- a/jupyterlab_comments/jupyterlab_comments/git_commands.py
+++ b/jupyterlab_comments/jupyterlab_comments/git_commands.py
@@ -46,7 +46,13 @@ class Git(Configurable):
 
 	def push_local_comments(self, git_root_dir):
 		self.run(git_root_dir, 'appraise', 'pull', self.remote)
-		self.run(git_root_dir, 'appraise', 'push', self.remote)
+		return_code = subprocess.call(['git', 'appraise', 'push', self.remote], cwd=git_root_dir)
+		#Retry once if refs fail to get pushed
+		if return_code == 1:
+			self.run(git_root_dir, 'appraise', 'pull', self.remote)
+			self.run(git_root_dir, 'appraise', 'push', self.remote)
+
+
 
 
 	def inside_git_repo(self, path_to_file):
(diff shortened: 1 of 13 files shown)


jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -41,6 +41,15 @@ def status(self, repo_path):
 	def appraise_pull(self, git_root_dir):
 		self.run(git_root_dir, 'appraise', 'pull', self.remote)
 
+	def appraise_push(self, git_root_dir):
+		self.run(git_root_dir, 'appraise', 'push', self.remote)
+
+	def push_local_comments(self, git_root_dir):
+		self.run(git_root_dir, 'push', self.remote)
```
ojarjur, line 48: I do not think we should be automatically pushing the user's local commits as part of this handler. Merely running `git appraise push` should be sufficient.

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -93,8 +106,13 @@ def get_code_review_comments(self, file_path_from_repo_root, git_root_dir):
 		return review_json
 
 
-	def add_comment(self, file_path, server_root):
-		pass
+	def add_detached_comment(self, file_path_from_repo_root, git_root_dir, comment_string):
+		self.run(git_root_dir, 'appraise', 'comment', '-d', '-m', comment_string, '-f', file_path_from_repo_root)
+		self.push_local_comments(git_root_dir)
```
ojarjur, line 111: You're calling both `push_local_comments` here, and calling `appraise_push` in the `AddDetachedCommentsHandler.push` method after it invokes this method.

That means there is a redundant push, and one of them can be eliminated.

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -41,6 +41,20 @@ def status(self, repo_path):
...
 		self.run(git_root_dir, 'appraise', 'pull', self.remote)
 
+	def appraise_push(self, git_root_dir):
+		self.run(git_root_dir, 'appraise', 'push', self.remote)
+
+	def push_local_comments(self, git_root_dir):
+		self.run(git_root_dir, 'appraise', 'pull', self.remote)
+		return_code = subprocess.call(['git', 'appraise', 'push', self.remote], cwd=git_root_dir)
+		#Retry once if refs fail to get pushed
+		if return_code == 1:
```
gogasca, line 51: do we always get return code = 1? or we could get another non-zero value?

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -78,6 +92,10 @@ def get_code_review_comments(self, file_path_from_repo_root, git_root_dir):
 		"""
 		review_string = self.run(git_root_dir, 'appraise', 'show',
 									'-json')
+		if review_string is None:
+			print('No open code reviews')
+			return None
+
 		review_json = json.loads(review_string)
 		if not review_json:
```
gogasca, line 100: not sure if we will hit this section if review_string is already None above. Line 95.

jupyterlab_comments/jupyterlab_comments/git_commands.py
mkalil: thanks for pointing this out! I'm changing this so it catches all non-zero return code values

jupyterlab_comments/jupyterlab_comments/git_commands.py
mkalil: with git appraise, I believe review_string will be None if there are no open code reviews, but it will be an empty string if there is an open code review but with no comments on that code review yet. I added the first check for None so I can safely call json.loads(review_string)

ojarjur: Reviews have more than just the comments, so if you want to explicitly check if there are no comments (or no matching comments), then we need to check that the comment threads field is missing or empty.

ojarjur: Although, taking another look, this is really just a safeguard, and I think it is fine for it to be overly cautious by checking for a situation that might not happen.

The checks for no comments is done well in the subsequent code and strictly speaking the `review_json` field could still be falsey even if the string read in was not `None`.

PR #146, opened 2020-07-23
[Bigquery] Add version 1.1 to @jupyter-widgets/jupyterlab-manager install

cbwilkes:


README.md
```diff
@@ -50,7 +50,7 @@ environment.
    JupyterLab environment in development mode. Afterwards, `cd` back to the root
    of the repository.
 
-   - Run `jupyter labextension install @jupyter-widgets/jupyterlab-manager` if your extension requires ipywidget. Currently, this is a requirement for the jupyterlab_bigquery extension.
+   - Run `jupyter labextension install @jupyter-widgets/jupyterlab-manager@1.1` if your extension requires ipywidget. Currently, this is a requirement for the jupyterlab_bigquery extension.
```
prodonjs, line 53: So it sounds like adding the [@jupyter-widgets/jupyterlab-manager](https://www.npmjs.com/package/@jupyter-widgets/jupyterlab-manager) package dependency to the BigQuery extension's package.json doesn't work?

MRzNone: I tried installing through npm, but it wasn't working.  I think Jupyter has a special space to save its extension.

prodonjs: That's unfortunate. There's limited documentation on [companion packages](https://jupyterlab.readthedocs.io/en/stable/developer/extension_dev.html#companion-packages), but I agree it doesn't seem like there is any way to explicitly say one extension depends on another being installed

PR #147, opened 2020-07-23
[BigQuery] Create pinning dialog

hwing777:
Create pinning button and dialog. Currently has basic error displays in the data tree. Next PR will have a snackbar component added to display errors. Testing will also be included in next PR.


jupyterlab_bigquery/jupyterlab_bigquery/list_items_handler/service.py
```diff
@@ -144,3 +144,12 @@ def search_projects(self, search_key, project_id):
         })
 
     return {'results': fetched_results}
+
+  def get_project(self, project_id):
```
cbwilkes, line 148: Can you add a test for this function?

PR #149, opened 2020-07-24
[HardwareConfiguration] Adding class to wrap around required functionality of the GCP AI Plat…

### LINKED ISSUE #129: Develop service layer for Notebooks API methods
The following methods of the Notebooks API should be wrapped by a service that can call each operation, poll the long-running operation that is returned, handle any error condition, and proceed to the next sequence in the chain of operations.

- [stop](https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1beta1/projects.locations.instances/stop)
- [setMachineType](https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1beta1/projects.locations.instances/setMachineType)
- [setAccelerator](https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1beta1/projects.locations.instances/setAccelerator)
- [start](https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1beta1/projects.locations.instances/start)

The [_pollOperation](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/b183f9f272cd783118206072b7ad1bde47b5134b/jupyterlab_gcpscheduler/src/service/project_state.ts#L442) method inside the ProjectState class from the GCP Scheduler extension is a good implementation for polling a long-running operation.

yaminahq:
…form Notebooks API

Resolves #129 

This PR:
1. Adds the NotebookInstanceServerClass which wraps around the [Notebooks API](https://cloud.google.com/ai-platform/notebooks/docs/reference/rest) to access the stop, start, setMachineType and setAccelerator methods
2. Adds support for PATCH requests in the ProxyHandler class in the shared request handlers file as needed for some of the methods accessed
3. Implements testing for the NotebookInstanceServerClass

- Some of the methods I copied over from the gcpscheduler code, I can move them to a shared location once this is good to go.
- It also sets up some functions that can be used to add additional logic to check if the Notebooks API is enabled for a particular gcp project and enable it if its not

I'll bump up the version number of the shared module once this is ready to push.

### DIFF
--- a/jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
+++ b/jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
@@ -1,0 +1,451 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-disable @typescript-eslint/camelcase */
+
+import {
+  handleApiError,
+  TransportService,
+  POST,
+  PATCH,
+  InstanceMetadata,
+  getMetadata,
+} from 'gcp_jupyterlab_shared';
+
+const POLL_INTERVAL = 5000;
+const POLL_RETRIES = 3;
+
+interface Service {
+  endpoint: string;
+  name: string;
+  documentation: string;
+  isOptional: boolean;
+}
+
+interface AcceleratorConfig {
+  type: string;
+  coreCount: string;
+}
+
+export interface Instance {
+  name: string;
+  state?: string;
+  machineType?: string;
+  acceleratorConfig?: AcceleratorConfig;
+}
+
+/** Service wrapper state corresponding to a Notebook Instance. */
+export interface State {
+  serviceEnabled: boolean;
+  projectId: string;
+  instanceName: string;
+  locationId: string;
+}
+
+/** API service status */
+export interface ServiceStatus {
+  service: Service;
+  enabled: boolean;
+}
+
+export const NOTEBOOKS_API_SERVICE: Readonly<Service> = {
+  name: 'AI Platform Notebooks API',
+  endpoint: 'notebooks.googleapis.com',
+  documentation: 'https://cloud.google.com/ai-platform/notebooks/',
+  isOptional: false,
+};
+
+type Operation = gapi.client.servicemanagement.Operation;
+type ListServicesResponse = gapi.client.servicemanagement.ListServicesResponse;
+
+const SERVICE_MANAGER = 'https://servicemanagement.googleapis.com/v1';
+export const NOTEBOOKS_API_PATH = 'https://notebooks.googleapis.com/v1beta1';
+
+const GOOGLE_API_UNAUTHENITCATED_CODE = 16;
+
+export function isUnauthorized(err: any) {
+  // Check for Google API Error structure and Unauthenticated error code
+  // https://cloud.google.com/apis/design/errors#error_codes
+  if (err.result && err.result.error) {
+    return err.result.error.code === GOOGLE_API_UNAUTHENITCATED_CODE;
+  }
+
+  return false;
+}
+
+/**
+ * Wraps around the GCP AI Platform Notebooks API to manage notebook resources in Google Cloud.
+ * https://cloud.google.com/ai-platform/notebooks/docs/reference/rest
+ */
+export class NotebookInstanceServiceLayer {
+  private projectIdPromise?: Promise<string>;
+  private instanceNamePromise?: Promise<string>;
+  private locationIdPromise?: Promise<string>;
+  private metadataPromise?: Promise<InstanceMetadata>;
+
+  constructor(
+    private _transportService: TransportService,
+    private _authToken: string,
+    projectId?: string,
+    instanceName?: string,
+    locationId?: string
+  ) {
+    if (projectId) {
+      this.projectIdPromise = Promise.resolve(projectId);
+    }
+    if (instanceName) {
+      this.instanceNamePromise = Promise.resolve(instanceName);
+    }
+    if (locationId) {
+      this.locationIdPromise = Promise.resolve(locationId);
+    }
+  }
+
+  get projectId() {
+    return this._getProject();
+  }
+
+  set projectId(projectIdPromise: string | Promise<string>) {
+    if (typeof projectIdPromise === 'string') {
+      this.projectIdPromise = Promise.resolve(projectIdPromise);
+    } else {
+      this.projectIdPromise = projectIdPromise;
+    }
+  }
+
+  get instanceName() {
+    return this._getInstance();
+  }
+
+  set instanceName(instanceNamePromise: string | Promise<string>) {
+    if (typeof instanceNamePromise === 'string') {
+      this.instanceNamePromise = Promise.resolve(instanceNamePromise);
+    } else {
+      this.instanceNamePromise = instanceNamePromise;
+    }
+  }
+
+  get locationId() {
+    return this._getLocation();
+  }
+
+  set locationId(locationIdPromise: string | Promise<string>) {
+    if (typeof locationIdPromise === 'string') {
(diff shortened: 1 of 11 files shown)


jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
```diff
@@ -0,0 +1,453 @@
...
+
+  set transportService(transportService: TransportService) {
+    this._transportService = transportService;
+  }
+
+  set authToken(authToken: string) {
+    this._authToken = authToken;
+  }
+
+  /**
```
prodonjs, line 161: We can make the assumption that the Notebooks API will always be enabled for a user, unlike Scheduler, which relied on other services that are not directly related to CAIP Notebooks.

jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
```diff
@@ -0,0 +1,453 @@
...
+      console.error('Unable to determine project status', err);
+      throw err;
+    }
+  }
+
+  /**
+   * Enables the service specified returning a Promise for the
+   * operation.
+   */
+  async enableService(serviceName: string): Promise<Operation> {
```
prodonjs, line 191: As mentioned above, this is not necessary given that a user of a CAIP Notebook will have to already have enabled the API.

jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
```diff
@@ -0,0 +1,453 @@
...
+  }
+
+  return false;
+}
+
+/**
+ * Wraps around the GCP AI Platform Notebooks API to manage notebook resources in Google Cloud.
+ * https://cloud.google.com/ai-platform/notebooks/docs/reference/rest
+ */
+export class NotebookInstanceServiceLayer {
```
prodonjs, line 93: I'd consider shortening this to something like NotebooksService and renaming the file accordingly.

jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
```diff
@@ -0,0 +1,453 @@
...
+    } catch (err) {
+      console.error('Unable to stop notebook instance');
+      handleApiError(err);
+    }
+  }
+
+  /**
+   * Starts the notebook instance.
+   */
+  async start(): Promise<Instance> {
```
prodonjs, line 373: The code for this method is virtually identically to the code for `stop`. I'd consider having each method call a private method with a parameter typed as `start|stop` to affect the `path` value since that's the only difference.

jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
```diff
@@ -0,0 +1,453 @@
...
+      const [projectId, instanceName, locationId] = await Promise.all([
+        this.projectId,
+        this.instanceName,
+        this.locationId,
+      ]);
+      const name = `projects/${projectId}/locations/${locationId}/instances/${instanceName}`;
+      const pendingOperation = await this._transportService.submit<Operation>({
+        path: `${NOTEBOOKS_API_PATH}/${name}:stop`,
+        method: POST,
+        headers: { Authorization: `Bearer ${this._authToken}` },
```
prodonjs, line 357: This may be problematic. If you're using the `ServerProxyTransportService`, it's going to automatically add its own `Authorization` header based on the machine's credential. 

The [ClientTransportService](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/2b51a62b78c549bba72668e181acdccf0f217acc/shared/src/service/transport.ts#L98) is a better choice, but it forces the user to go through the OAuth flow, which we don't want. We may need to modify that class to allow an OAuth token to be passed in, which would bypass the initialization flow.

yaminahq: Ah I noticed that and was going to ask you how this the authorization plays out in our use case, since while testing it did seem to work. We can talk more about this during stand-up but it's a good call to use ClientTransportService as you suggested - I'll add that!

jupyterlab_gcedetails/src/service/notebook_instance_service_layer.ts
yaminahq: Since the setAccelerator and setMachineType are also very similar apart from the request path, request method, request body and error, would it be good practice to have these methods use the private method as well or would that be too much? I could then just pass an error message and request object to the private buildRequest method and it may reduce duplicated code a bit more

prodonjs: I'd probably say it's worth factoring those two methods to use some common method as well since each of them return a long-running operation that eventually resolves to an `Instance`.

shared/src/service/transport.ts
```diff
@@ -36,6 +36,7 @@ export interface ApiResponse<T> {
 export interface TransportService {
   /** Submit the Google API request and resolve its response. */
   submit<T>(request: ApiRequest): Promise<ApiResponse<T>>;
+  setAccessToken?(accessToken: string): Promise<void>;
```
prodonjs, line 39: Since this is only applicable to the ClientTransportService, I would suggest that you don't make it part of the general TransportService interface and instead add it specifically to the ClientTransportService. Then, in the NotebookService, you can explicitly declare the constructor as requiring a ClientTransportService object.

shared/src/service/transport.ts
```diff
@@ -101,13 +102,25 @@ export class ClientTransportService implements TransportService {
...
   private authenticated = false;
 
-  constructor(private clientId: string) {}
+  /**
+   * If useAccessToken is set, you will need to call setAccessToken on the
+   * ClientTransportService instance before making your first request and
+   * call it again whenever the token expires
+   */
+
+  constructor(private clientId: string, private useAccessToken?: boolean) {}
```
prodonjs, line 111: I'm thinking that the signature here might need to be different. For one, I don't know that you need a flag for useAccessToken. You could simply have an accessToken private data member and a setter for it. The presence of a value would be the cue to avoid doing the authentication flow.

Also, clientId would be optional, since if you're providing the auth token, you don't need a clientId to do the OAuth flow with, and in our case, we won't have one.

jupyterlab_gcedetails/src/service/notebooks_service.ts
```diff
@@ -340,114 +275,102 @@ export class NotebookInstanceServiceLayer {
...
+      console.error(errorMessage);
       handleApiError(err);
     }
   }
 
+  /**
+   * Stops the notebook instance.
+   */
+  async stop(): Promise<Instance> {
+    const [projectId, instanceName, locationId] = await Promise.all([
```
prodonjs, line 300: The block that gets the instance name is repeated in each method. Sorry if my comments in Chat were not clear but this is a candidate for something that I would put into a private method since it is repeatedly used.

jupyterlab_gcedetails/src/service/notebooks_service.ts
```diff
@@ -340,114 +275,102 @@ export class NotebookInstanceServiceLayer {
     return this.metadataPromise;
   }
 
-  /**
-   * Stops the notebook instance.
-   */
-  async stop(): Promise<Instance> {
+  private async submitRequest(
```
prodonjs, line 278: :)

shared/src/service/transport.ts
```diff
@@ -101,13 +102,25 @@ export class ClientTransportService implements TransportService {
...
 
   async submit<T>(request: ApiRequest): Promise<ApiResponse<T>> {
     await this.initializeClient();
     return gapi.client.request(request);
   }
 
+  async setAccessToken(accessToken: string): Promise<void> {
+    await this.initializeClient();
```
prodonjs, line 119: See my previous comment. I wouldn't have the GAPI initialization here. I would just have it be a dumb setter (you could use the ES6 `set` style) for a private data member. The initialization flow would still occur at the time a request is made, where the GAPI library would be loaded. If the accessToken private data member was set, then it skips the next step.

shared/src/service/transport.ts
yaminahq: Mainly had it here for testing purposes - I'll change it!

shared/src/service/transport.ts
```diff
@@ -172,5 +183,10 @@ export class ClientTransportService implements TransportService {
         this.authenticatedPromise = undefined;
       }
     }
+
+    if (this._accessToken) {
```
prodonjs, line 187: As I suggested in Chat, I would call this in the `submit` method right before the call to `gapi.client.request` so you can be sure that you use the most recently provided accessToken value.

prodonjs: Ahh my apologies. With the way the diff viewer collapsed the common code, I couldn't see that this was calling `gapi.client.setToken` irregardless if any of the other conditional blocks were true. This accomplishes what I was mentioning in the previous comment.

prodonjs:
@yaminahq Can you bump the version for the shared package (both the Python and TypeScript)? That will be necessary since you updated code there.

PR #150, opened 2020-07-24
[BigQuery] Table preview and query results fixes

CynthiaXJia:
Fixes previously incorrectly shown empty records by fully populating table cells with empty values. For now, shows repeated fields as a string, which truncates if too long. Also formats null values in the table so they show up as 'null' instead of empty.


jupyterlab_bigquery/jupyterlab_bigquery/details_handler/details_handler.py
```diff
@@ -54,100 +55,158 @@ def get_dataset_details(client, dataset_id):
...
 def format_preview_field(formatted_fields, field, field_name):
   if field.field_type == 'RECORD':
-    for record_entry in field.fields:
-      format_preview_field(formatted_fields, record_entry, field_name + field.name + ".")
+    if field.mode == 'REPEATED':
+      formatted_fields.append(field.name)
+    else:
+      for record_entry in field.fields:
+        format_preview_field(formatted_fields, record_entry,
+                             field_name + field.name + ".")
```
CynthiaXJia, line 135: Repeated fields that are records won't get expanded out into multiple columns for each subfield anymore, since their repeated items will just be listed as a string.

PR #152, opened 2020-07-24
[BigQuery] syntax debug highlight

MRzNone:
# Syntax Error
image

# Not found error

image

### DIFF
--- a/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
+++ b/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
@@ -8,7 +8,7 @@ import {
   QueryId,
 } from '../../../reducers/queryEditorTabSlice';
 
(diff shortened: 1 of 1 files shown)


MRzNone:
Monaco supports display error message when hovering over the error mark, but the react version we are using is not supporting is.  I created an [issue](https://github.com/suren-atoyan/monaco-react/issues/93) on their side.

I decided to keep the debug message below since the message display on hovering is not available.

suren-atoyan:
I just want to inform you that the react version of the editor also supports it. The problem is there was a typo in your code, I've already sent the correct version; check it [here](https://github.com/suren-atoyan/monaco-react/issues/93).

MRzNone:
Fixed typo.  Hovering works.  Removed dry-run error message.

# Syntax error
image


# "semantic" error
image

cbwilkes:
Is dry-run error message still removed. I don't think we should remove the print out and force the user to hover the mouse.

cbwilkes:
We will add back the dry run error message in a later PR

PR #153, opened 2020-07-26
[WIP] Feat/authentication token retrieval

l1ghtspeed:
Up for preliminary feedback, everything is manually tested and working as expected


.gitignore
```diff
@@ -60,6 +60,9 @@ yarn.lock
 # dotenv environment variables file
 .env
 
+# json secrets
+*.secrets.json
```
l1ghtspeed, line 64: I have the secrets stored in a json file, but I can also change it to use a .env instead

jupyterlab_gcedetails/auth_token_server/app.js
```diff
@@ -0,0 +1,73 @@
...
+  keys.client_id,
+  keys.client_secret,
+  keys.redirect_uris[0]
+);
+
+const scopes = ['https://www.googleapis.com/auth/compute'];
+
+google.options({ auth: oauth2Client });
+
+const errorFilePath = path.join(__dirname, 'error.html');
```
l1ghtspeed, line 33: Some rudimentary error handling, is there a specific way that we should do it? It's just kind of weird since we're not using typical http error codes

jupyterlab_gcedetails/auth_token_server/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "private": true,
```
l1ghtspeed, line 2: Not sure if there was a specific format to these, I just followed the one in the main folder

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -0,0 +1,56 @@
...
+      });
+    } else {
+      this.setState({
+        status: Status['Reshaping Instance'],
+        authToken: message.data['credentials'],
+      });
+    }
+  };
+  private readonly oAuthHost =
+    'https://jupyterlab-interns-sandbox.uc.r.appspot.com/authorize';
```
l1ghtspeed, line 32: Are we going to keep the token server hosted on this url?

jupyterlab_gcedetails/auth_token_server/app.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Create a new OAuth2 client with the configured keys.
+ */
+const oauth2Client = new google.auth.OAuth2(
+  keys.client_id,
+  keys.client_secret,
+  keys.redirect_uris[0]
+);
+
+const scopes = ['https://www.googleapis.com/auth/compute'];
```
l1ghtspeed, line 29: I wasn't able to find a scope for specifically the notebooks api

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+
+enum Status {
```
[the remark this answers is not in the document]
l1ghtspeed: The token retrieval flow doesn't actually render anything, maybe I could just split it out into a function in a different file?

jupyterlab_gcedetails/auth_token_server/error.html
```diff
@@ -0,0 +1,9 @@
+<html>
```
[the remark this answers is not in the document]
prodonjs: Nevermind, I didn't realize it was sending back an error as we discussed in yesterday's standup.

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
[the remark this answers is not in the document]
prodonjs: Yeah I think that is reasonable. The main thing I would do is separate the code that is responsible for opening the popup and listening for the Auth token message from the part that is going to be responsible for initiating the API calls to do the reshaping and polling their statuses. Combining those two behaviors into a single component will make for a relatively difficult testing experience.

jupyterlab_gcedetails/auth_token_server/app.js
```diff
@@ -23,10 +23,10 @@ if (fs.existsSync(keyPath)) {
 const oauth2Client = new google.auth.OAuth2(
   keys.client_id,
   keys.client_secret,
-  keys.redirect_uris[0]
+  'https://jupyterlab-interns-sandbox.uc.r.appspot.com/oauth2callback'
```
prodonjs, line 26: *nit*, but it's usually a good practice to support local development and production with some sort of logic based on environment variables. So you could something like if NODE_ENV === 'dev' use localhost, otherwise prod.

PR #157, opened 2020-07-27
[BigQuery] Search all pinned projects

hwing777:
Change searching to search all pinned projects in the data tree. Update the 'enable Data Catalog API' dialog to appear when searching returns an error.

### DIFF
--- a/jupyterlab_bigquery/src/components/list_items_panel/list_tree_panel.tsx
+++ b/jupyterlab_bigquery/src/components/list_items_panel/list_tree_panel.tsx
@@ -19,7 +19,7 @@ import { WidgetManager } from '../../utils/widgetManager/widget_manager';
 import ListSearchResults from './list_search_results';
 import { QueryEditorTabWidget } from '../query_editor/query_editor_tab/query_editor_tab_widget';
 import { updateDataTree, addProject } from '../../reducers/dataTreeSlice';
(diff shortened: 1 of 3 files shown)


jupyterlab_bigquery/src/components/list_items_panel/list_tree_panel.tsx
```diff
@@ -305,10 +318,10 @@ class ListItemsPanel extends React.Component<Props, State> {
                 style={{ color: 'blue' }}
                 href="https://console.developers.google.com/apis/api/datacatalog.googleapis.com/overview"
               >
-                Google Data Catalog API.
+                Google Data Catalog API
               </a>{' '}
-              Once you click "Enable", this may take up to 2-3 minutes before
-              you can start searching.
+              for all pinned projects. Once you click "Enable", this may take up
```
cbwilkes, line 323: Maybe state `this may take several minutes` instead? Can we confidently promise an upper bound?

hwing777: This is the same message that is in API-enabling popups from the cloud console. Should I change it to several minutes or keep what is in the webUI to maintain consistency?

cbwilkes: We can leave it be for consistency

PR #162, opened 2020-07-28
Handler tests

mkalil:


jupyterlab_comments/jupyterlab_comments/handlers_test.py
```diff
@@ -0,0 +1,233 @@
...
+import tempfile
+import handlers
+import subprocess
+import unittest
+import json
+import tornado.ioloop
+import tornado.web
+import tornado.testing
+
+remote_dir = tempfile.TemporaryDirectory()
```
ojarjur, line 26: We aren't cleaning up the temporary directories anywhere, which means they will stick around until the machine running the test reboots.

We should take advantage of the fact that these return context managers so that we use them in a `with` statement. That way, the temporary director(y|ies) get automatically deleted after the `with` statement completes.

If you want to use multiple directories you can do this like:

```py
    with tempfile.TemporaryDirectory() as remote_dir, \
           tempfile.TemporaryDirectory() as not_get_dir, \
           tempfile.TemporaryDirectory() as dir1, \
           tempfile.TemporaryDirectory() as dir2:
       ...
```

Alternatively, you can use a single temporary directory as a parent and create all the other directories as regular directories under that:

```py
    with tempfile.TemporaryDirectory() as working_dir:
       remote_dir = os.path.join(working_dir, "remote")
       os.makedirs(remote_dir)
       not_git_dir = os.path.join(working_dir, "not_git")
       os.makedirs(not_git_dir)
       dir1 = os.path.join(working_dir, "dir1")
       os.makedirs(dir1)
       dir2 = os.path.join(working_dir, "dir2")
       os.makedirs(dir2)
       ... code to run the tests ...
```

You could wrap these `with` statements around the code to the unittest `run` method, but I think it would be better to use this inside of each individual test method, so that they do not reuse test repositories between different tests.

So, if you combined that with the suggested changes to the `setup_repos` method below, you would wind up with each test method looking something like:

```py
    def test_...(self)
        with tempfile.TemporaryDirectory() as working_dir:
            test_env = setup_repos(working_dir)
            ... code to perform the test...
```

jupyterlab_comments/jupyterlab_comments/handlers_test.py
```diff
@@ -0,0 +1,233 @@
...
+dir1_path = dir1.name
+dir2_path = dir2.name
+
+
+home = os.path.expanduser("~")
+FNULL = open(os.devnull, 'w')
+
+def create_files(num_files, dir_name):
+    for i in range(1, num_files + 1):
+        with open(os.path.join(dir_name, "test" + str(i) + ".py"), 'a') as temp_file:
```
ojarjur, line 39: Calling this `temp_file` confused me at first as I was thinking that it was a file that only existed in the scope of the `with` statement (which is not the case).

I would suggest renaming this to something like `test_file`.

jupyterlab_comments/jupyterlab_comments/handlers_test.py
```diff
@@ -0,0 +1,233 @@
...
+    app = tornado.web.Application()
+    host_pattern = '.*$'
+    app.add_handlers(host_pattern, [
+        ('/detachedComments', handlers.DetachedCommentsHandler),
+        ('/remotePull', handlers.PullFromRemoteRepoHandler),
+        ('/addDetachedComment', handlers.AddDetachedCommentHandler),
+        ('/reviewComments', handlers.ReviewCommentsHandler)])
+    return app
+
+def setup_repos():
```
ojarjur, line 52: I would suggest passing in the current working directory as an argument (which could be a temp directory), and then creating all of the repo dirs as subdirectories of that.

That way, you can have a different set of repositories used by each test.

I would additionally suggest returning the names of the repository directories so that they don't have to be hard-coded.

That would be easiest if you defined an object that represented a test environment, made each of the created repository names a field in that object, and then returned the object.

Something like:

```py

class TestEnv(object):
    def __init__(self, remote, clone1, clone2):
        self.remote = remote
        self.clone1 = clone1
        self.clone2 = clone2

def setup_repos(cwd):
   ...
   return TestEnv(remote_dir, dir1, dir2)
```

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -29,7 +31,7 @@ def __init__(self, **kwargs):
 
 	def run(self, cwd, *args):
 		try:
-		  return subprocess.check_output(['git'] + list(args), cwd=cwd)
+		  return subprocess.check_output(['git'] + list(args), cwd=cwd, stderr=subprocess.DEVNULL)
```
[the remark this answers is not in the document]
mkalil: Running the tests with multiple git calls was producing a lot of output that I didn't want to see every time I re-ran them -- I will take this out since it is needed for debugging the actual extension

jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -16,6 +16,8 @@
 import json
 from traitlets.config.configurable import Configurable
 from traitlets import Int, Float, Unicode, Bool
+import os
+FNULL = open(os.devnull, 'w')
```
gogasca, line 20: New line between imports + FNULL

ojarjur:
This looks good to me besides the comments I noted above.

PR #167, opened 2020-07-29
[uCAIP] Model testing feature with code generation

josie-garza:
- list endpoints that the model is deployed on or gives option to deploy model
- ability to copy endpoint ID and undeploy model from endpoint list
- ability to predict from the UI (includes placeholder values with correct format)
- ability to generate python code to easily copy to a notebook
- if model is undeployed from extension managed endpoint, the endpoint will be deleted

image
image

The button padding issues in the images will be fixed with shared version 1.0.5


jupyterlab_automl/jupyterlab_automl/service.py
```diff
@@ -266,3 +277,66 @@ def create_dataset_from_file(self, display_name, file_name, file_data):
...
+    return endpoints
+
+  def check_deploying(self, model_name):
+    endpoints = self._endpoint_client.list_endpoints(parent=self._parent).endpoints
+    name = "ucaip-extension/" + model_name
+    filtered = filter(lambda x: name == x.display_name, endpoints)
+    filtered = list(filtered)
+    if len(filtered) > 0:
+      return [self._build_endpoint("None", filtered[0])]
+    return []
```
josie-garza, line 318: This isn't my favorite thing but basically just checking to see if an extension managed endpoint for this model exists without a deployed model meaning that the model is currently deploying on that endpoint. I couldn't really think of a better way to do this. The alternative would be to continue to show the "deploy model" button while the model is deploying which could lead to duplicate deployments.

PR #169, opened 2020-07-29
[GitSync] initial commit using cookiecutter template

ashleyswang:
Initial commit for jupyterlab_gitsync project using cookiecutter template from GCP.


ashleyswang:
@googlebot I fixed it.

PR #174, opened 2020-07-30
[Optimizer] add conditional parameters tree view for study details

kpfromer:
# Description

- Adds a tree view for conditional parameters in the study details page.
- When a user clicks on a node it opens a details dialog about that parameter.

# Images
image
image


jupyterlab_caip_optimizer/src/components/create_study.spec.tsx
```diff
@@ -33,3 +39,315 @@ describe('Dropdown list', () => {
     expect(dropdown).toEqual(goalDropdownList);
   });
 });
+
+describe('Create Study Page', () => {
+  it('creates a study and redirects to the study details page', async () => {
```
kpfromer, line 44: This is really strange. I am pretty sure that these tests are already on master.

jupyterlab_caip_optimizer/src/components/study_details/index.spec.tsx
```diff
@@ -0,0 +1,140 @@
+import * as React from 'react';
```
kpfromer, line 1: Same here.

jupyterlab_caip_optimizer/style/index.css
```diff
@@ -5,3 +5,37 @@
   background-image: url('./images/storage_24px.svg');
   background-size: contain;
 }
+
```
kpfromer, line 8: Should I move these styles out of this file?

jupyterlab_caip_optimizer/src/index.ts
```diff
@@ -58,8 +58,6 @@ const createManagedWidget = <
 };
 
 async function activate(app: JupyterFrontEnd) {
-  store.dispatch(fetchStudies());
```
kpfromer, line 61: This removed a useless API call since the user has no metadata at this point.

jupyterlab_caip_optimizer/style/index.css
angelakuo: That SGTM if you have a more specific place to put them

jupyterlab_caip_optimizer/src/components/parameter_details_dialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <>
+          <DialogTitle>Parameter "{spec.parameter}"</DialogTitle>
+          <DialogContent>
+            <DialogContentText>
+              The study config parameter specification.
+            </DialogContentText>
+            <ParameterDetails spec={spec} />
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={onClose} color="primary" variant="contained">
```
angelakuo, line 35: I think Exit / Cancel buttons are generally not primary in color (only action buttons that you want to direct attention to and will do stuff are raised/colored)

jupyterlab_caip_optimizer/src/components/parameter_details_dialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+      open={!!spec}
+      onClose={onClose}
+      data-testid="parameterDetailsDialog"
+    >
+      {!!spec && (
+        <>
+          <DialogTitle>Parameter "{spec.parameter}"</DialogTitle>
+          <DialogContent>
+            <DialogContentText>
+              The study config parameter specification.
```
angelakuo, line 30: I feel like this text isn't necessary, we could probably do without it

jupyterlab_caip_optimizer/src/components/parameter_details_dialog.tsx
kpfromer: Should I then remove the contained variant?

kpfromer:
Changes:
image

PR #178, opened 2020-07-31
[HWCONFIG]/integrate api calls

l1ghtspeed:
Currently a few of the snapshots are failing due to the presence of the extra icon, will update before merge. Demo link: https://screencast.googleplex.com/cast/NjQ0MTMwMTQyOTk3NzA4OHw3NjQ5MmZiMC1iMQ


jupyterlab_gcedetails/src/components/auth_token_retrieval.tsx
```diff
@@ -1,14 +1,20 @@
-export function AuthTokenRetrieval(callback) {
+export function AuthTokenRetrieval(callback: (err, credentials) => void) {
```
prodonjs, line 1: *nit* Function names should begin with a lowercase letter.

jupyterlab_gcedetails/src/components/auth_token_retrieval.tsx
```diff
@@ -1,14 +1,20 @@
...
-    authPopup.close();
-    window.removeEventListener('message', listener);
-    callback(message.data['error'], message.data['credentials']);
+    if (
+      message.origin === authOrigin &&
+      (message.data['error'] || message.data['credentials'])
+    ) {
+      authPopup.close();
+      window.removeEventListener('message', listener);
+      callback(message.data['error'], message.data['credentials']);
```
prodonjs, line 12: What do you think about writing this as a Promise returning function? It could return `Promise<string>`. When the message was received from the auth site, it would check if `data.error` was set, and if not, call the `.resolve()` method on the promise. Otherwise, it would call `.reject()` with the error that was returned. Then, you could use `async/await` in the component that uses this function.

That would be consistent with how most of the other APIs we interact with work.

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -0,0 +1,173 @@
...
+  'Authorizing' = 0,
+  'Stopping Instance' = 1,
+  'Reshaping Instance' = 2,
+  'Starting Instance' = 3,
+  'Complete' = 4,
+  'Error' = 5,
+}
+
+const statusInfo = [
+  'Getting authorization token to reshape machine. A dialog box should have popped up.',
```
prodonjs, line 58: What do you think about "Please complete the OAuth 2.0 authorization steps in the popup" for the second sentence?

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+export class HardwareScalingStatus extends React.Component<Props, State> {
+  private notebookService: NotebooksService;
+  constructor(props: Props) {
+    super(props);
+    this.state = {
+      status: Status.Authorizing,
+    };
+    const clientTransportService = new ClientTransportService();
+    this.notebookService = new NotebooksService(clientTransportService);
```
prodonjs, line 84: Instantiation in the constructor is generally an antipattern because it becomes almost impossible to unit test to mock out API calls.

Generally, you're going to want to create service classes that affect state and make API calls near the top-level of your application and then pass them down. Take a look at the GCP Scheduler where we create these classes at the time the extension is activated (https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/e4e1df377fc5310d15dbe8e68a542afb452b9fcd/jupyterlab_gcpscheduler/src/index.ts#L83).

It may be useful to write some kind of utility function or class to obtain this instance if we find that we need to pass the service down through too many component layers. Happy to chat about this when we meet tomorrow.

jupyterlab_gcedetails/src/components/auth_token_retrieval.tsx
```diff
@@ -1,21 +1,30 @@
...
+    try {
+      const listener = (message: MessageEvent) => {
+        if (
+          message.origin === authOrigin &&
+          (message.data['error'] || message.data['credentials'])
+        ) {
+          authPopup.close();
+          window.removeEventListener('message', listener);
+          if (message.data['error']) {
+            throw new Error();
```
prodonjs, line 15: To simplify and eliminate the need for the try/catch, I think you could also just call `reject(message.data['error'[)` and then return. I would suggest adding the error information to the rejection so you could potentially surface what failed.

PR #180, opened 2020-08-03
[HardwareConfiguration] Adding confirmation page before user submits VM reshaping request

### LINKED ISSUE #155: Add confirmation view to the HardwareScalingDialog
Before a user submits a VM reshaping request, we should ask them to confirm the selected configuration. We can also display pricing details here. The cloud console version of this looks like this:
Screenshot 2020-07-25 at 5 00 36 PM

yaminahq:
image

This commit:
- adds the confirmation page that appears in the hardware scaling dialog after the user submits the hardware scaling form
- modifies the server code to return gpu count as part of the details returned from nvidia-smi
- adds a HardwareConfiguration interface that represents a VM shape that the user can modify
- deletes some unused code (MASTER_TYPES)

Resolves #155

### DIFF
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
@@ -121,8 +121,6 @@ export class HardwareScalingForm extends React.Component<Props, State> {
     });
   }
 
-  // TODO: Implement submit functionality
-  //eslint-disable-next-line @typescript-eslint/no-empty-function
   private submitForm() {
     const configuration = { ...this.state.configuration };
     this.props.onSubmit(configuration);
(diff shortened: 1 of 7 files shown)


jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -87,41 +85,52 @@ export class HardwareScalingForm extends React.Component<Props, State> {
...
+        ...this.state.configuration,
+        gpuType: event.target.value,
+        gpuCount: event.target.value
+          ? (ACCELERATOR_COUNTS_1_2_4_8[0].value as string)
+          : '',
+      },
     });
   }
 
   // TODO: Implement submit functionality
```
prodonjs, line 124: I believe you can remove this TODO.

PR #181, opened 2020-08-03
[BigQuery] Resize editor on side panel open and close

CynthiaXJia:
Uses resize signals from the full window widget and [react-resize-detector](https://www.npmjs.com/package/react-resize-detector) in the in-cell element to resize the query editor with `editor.layout()`. 

I was hoping to implement the resizing in both types of editor (both in-cell and full window) with widget signals, but I couldn't get it to work properly with the in-cell editor. However, it seems that react-resize-detector is fairly widely used and it includes a polyfill so that the resizing is compatible in most web browsers, so I feel ok about using it.

I'm wondering if we should convert the widget signal resizing to also use react-resize-detector just for consistency? Currently there is a slight delay in the in-cell editor resizing at times, so you will see a ghost of the old size for a second before it resizes. This does not occur in the full-window editor, so this might be a reason to refrain from converting everything to react-resize-detector. However, would love to hear any thoughts on this.

### DIFF
--- a/jupyterlab_bigquery/src/components/query_editor/query_editor_incell/query_editor_incell.tsx
+++ b/jupyterlab_bigquery/src/components/query_editor/query_editor_incell/query_editor_incell.tsx
@@ -9,6 +9,7 @@ import {
   generateQueryId,
 } from '../../../reducers/queryEditorTabSlice';
 import { DOMWidgetView } from '@jupyter-widgets/base';
+import ReactResizeDetector from 'react-resize-detector';
 
 interface QueryEditorInCellProps {
   queries: { [key: string]: QueryResult };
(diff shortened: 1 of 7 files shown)


jupyterlab_bigquery/src/components/query_editor/query_editor_incell/query_editor_incell.tsx
```diff
@@ -40,19 +41,26 @@ export class QueryEditorInCell extends Component<QueryEditorInCellProps, {}> {
...
+              />
+              {showResult ? (
+                <QueryResults queryId={this.queryId} editorType="IN_CELL" />
+              ) : (
+                undefined
+              )}
+            </div>
+          );
+        }}
+      </ReactResizeDetector>
```
CynthiaXJia, line 63: Similar to `UseSignal`, `ReactResizeDetector` will rerender its children whenever the width changes and is passed down. There is also a HOC available, but I thought this was more similar to how we resize the full-window editor.

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
```diff
@@ -67,7 +69,6 @@ interface QueryRequestBodyType {
 
 const SQL_EDITOR_OPTIONS: editor.IEditorConstructionOptions = {
   lineNumbers: 'on',
-  automaticLayout: true,
```
CynthiaXJia, line 70: This was removed and replaced by a window resize listener, since this option actually calls a check every few milliseconds to see if the window was resized, which supposedly can slow the editor down.

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
```diff
@@ -245,6 +272,9 @@ class QueryTextEditor extends React.Component<
   }
 
   handleEditorDidMount(_, editor) {
+    if (this.editorRef.current) {
+      this.setState({ height: this.editorRef.current.clientHeight });
+    }
```
CynthiaXJia, line 277: The `editor.layou({width, height})` call needs to know a height along with a width, so here is where we read/set the current height in order to pass that into the resize dimensions.

PR #184, opened 2020-08-03
[BigQuery] open query editor from side panel

MRzNone:


MRzNone:
Fixed render race condition by delaying render of Monaco editor.

image

cbwilkes:
Seems like there are conflicts. Can you check those out?

PR #186, opened 2020-08-04
[Optimizer] Visualizations - Parallel Coordinates Feature

junhyeokkwak:
## Overview
Implements the foundation for the Optimizer's visualization page with parallel coordinates. This allows users to chart the trials on parallel axes indicating trial number, parameters, and final measurement metrics.

## Features
* Process through the trials and display only the completed trials with final measurements.
* Populate the sliders and axes using the trials and study configuration information.
* Plot the trials on the parallel coordinates axes.
* Click/drag and select a portion on one of the vertical axis to only view a subset of trials.
* Double-click anywhere on the graph to reset the selected range along the axis.
* Move the sliders below to zoom in/out on the axes.
* Click on one of the metric radio buttons to view the red-blue color spectrum.

## Screenshots
VizViz-F3-1
VizViz-F3-2
VizVizF3-5
VizVizF3-6


jupyterlab_caip_optimizer/src/components/graphs/parallel_coordinates.tsx
```diff
@@ -0,0 +1,395 @@
+import React, { useRef, useEffect } from 'react';
```
angelakuo, line 1: So I would probably make sure this file is well commented/documented for a smooth handoff

jupyterlab_caip_optimizer/src/components/graphs/parallel_coordinates.tsx
```diff
@@ -0,0 +1,395 @@
...
+                lineDataList[clickedAxisName] <
+                  Math.max(
+                    selectBoxDetails.initialValue,
+                    selectBoxDetails.currentValue
+                  )
+              )
+                withinRange = true;
+              break;
+            case 'DISCRETE':
+            case 'CATEGORICAL': {
```
angelakuo, line 277: nit: The { } don't seem necessary does it?

jupyterlab_caip_optimizer/src/components/suggest_trials/index.spec.tsx
```diff
@@ -1,3 +1,4 @@
+jest.setTimeout(100000);
```
angelakuo, line 1: Looks kind of strange as the first line, would this still work in a beforeEach block in the top level of this test? (same for the other file)

jupyterlab_caip_optimizer/src/components/graphs/parallel_coordinates.tsx
kpfromer: We get linting errors from eslint about not including { } when we define variables in the case statement. https://eslint.org/docs/rules/no-case-declarations

junhyeokkwak:
> Everything looks great so far! However the [lines going out of bounds](https://screenshot.googleplex.com/6NL5XxVaDUp.png) do look a bit strange (I remember you mentioning this). Do you think it's possible to just hide those? I think it's safe to assume that if the user did want to see them they would adjust the relevant parameters

I made the changes so that all the lines outside of the axes will just be hidden.

jupyterlab_caip_optimizer/src/components/graphs/parallel_coordinates.tsx
junhyeokkwak: done!

PR #191, opened 2020-08-05
[HardwareConfiguration] Bugfix/instance name split

l1ghtspeed:
A quick bugfix for the instance name parsing


jupyterlab_gcedetails/src/details_widget.tsx
```diff
@@ -115,7 +115,9 @@ export class VmDetails extends React.Component<Props, State> {
       this.setState({ details: details });
       notebookService.projectId = details.project.projectId;
       notebookService.locationId = details.instance.zone;
-      notebookService.instanceName = details.instance.name;
+      const instanceNameSplit = details.instance.name.split('/');
```
prodonjs, line 118: You can also use the `pop()` method for future cases like this where you want the last piece of a split-string. It mutates the array but in this case you don't care about the other pieces.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop

PR #193, opened 2020-08-05
[BigQuery] Refactor tree and implement refreshing

hwing777:
## Refactor Tree
Refactor the data tree in 'list_tree_item.tsx' and the search results in 'list_search_results.tsx'. The files now use class inheritance to be more self-contained and to prevent repetitive code between the files. Functionality of both files is identical to before refactoring (no features added).

## Refreshing
Create refreshing feature for the data tree in 'list_tree_panel.tsx'. Refreshing can be triggered for the whole panel, or for individual projects/datasets via their context menu.

image

### Minor UI Changes
Also change header styling in 'list_tree_panel.tsx' to hide button overflow text with ellipsis instead of wrapping. Add tooltips to display the full title of buttons even when overflowed.

image

### DIFF
--- a/jupyterlab_bigquery/src/components/list_items_panel/list_tree_item.tsx
+++ b/jupyterlab_bigquery/src/components/list_items_panel/list_tree_item.tsx
@@ -14,6 +14,8 @@ import {
   DataTree,
   Project,
   Dataset,
+  Table,
+  Model,
   ListDatasetsService,
   ListTablesService,
   ListModelsService,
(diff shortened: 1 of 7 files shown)


hwing777:
Apologies for big PR. Almost all changes in this diff are just refactoring to make things more self-contained, which was necessary to properly implement refreshing.

cbwilkes:
I think there is a conflict that needs to be resolved.

jupyterlab_bigquery/src/components/list_items_panel/list_tree_item.tsx
```diff
@@ -85,73 +87,88 @@ const localStyles = stylesheet({
...
-  const openModelDetails = (event, model) => {
+  openModelDetails = (event, model) => {
     event.stopPropagation();
     const service = new ModelDetailsService();
     const widgetType = ModelDetailsWidget;
-    context.manager.launchWidgetForId(
+    this.props.context.manager.launchWidgetForId(
       model.id,
       widgetType,
       service,
```
hwing777, line 174: Resolved conflict and added support for opening model details

PR #195, opened 2020-08-05
[Optimizer] create study conditional parameters

kpfromer:
# Description

- User can enable a conditional parameter
  - User then selects the parent parameter (only valid parameters are shown, ie can not be the current parameter, and must be a categorical, integer or discrete parameter spec)
   - User then selects the valid options that the parameter shows up for
- **Should I show the parameter tree (like the one on the study details page)?** It would be easy for me to add.

# Images

image
image
image
**Boom - Created study!**
image

### DIFF
--- a/jupyterlab_caip_optimizer/src/setupTests.ts
+++ b/jupyterlab_caip_optimizer/src/setupTests.ts
@@ -38,6 +38,21 @@ beforeAll(() => {
     }
     return originalError(...args);
   });
+
+  // https://github.com/mui-org/material-ui/issues/15726#issuecomment-493124813
+  // TODO: remove upgrading to jest v26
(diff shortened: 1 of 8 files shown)


jupyterlab_caip_optimizer/src/setupTests.ts
```diff
@@ -38,6 +38,21 @@ beforeAll(() => {
     }
     return originalError(...args);
   });
+
+  // https://github.com/mui-org/material-ui/issues/15726#issuecomment-493124813
+  // TODO: remove upgrading to jest v26
+  global.document.createRange = () => ({
```
kpfromer, line 44: This resolved some testing issues that came up due to animations with material ui and the autocomplete component.

PR #200, opened 2020-08-06
Added gpu graphs

l1ghtspeed:
image
Here's how it currently looks, let me know if there are any styling changes I should make.


jupyterlab_gcedetails/src/components/resource_utilization_charts.tsx
```diff
@@ -48,6 +55,9 @@ const STYLES = stylesheet({
   utilizationChartsContainer: {
     padding: '10px 20px 20px 0px',
   },
+  flexspan: {
```
prodonjs, line 58: I think you can take advantage of some of the stuff that's already been included with `typestyle` and `csstips` such as `...csstips.horizontal` if you want a horizontal flex layout.

PR #202, opened 2020-08-06
[HardwareConfiguration] Prepopulate scaling form with user's current configuration, adding in…

yaminahq:
…fo messages and modifying styling

Screenshot 2020-08-06 at 9 51 59 AM

Screenshot 2020-08-06 at 10 20 00 AM

This commit:
 - Prepopulates the hardware scaling form with the users current configuration and consequently modifies the machineType variables to be type MachineType in the state of the HardwareScalingForm and Option when used in the select
 - Adds a message informing the user of potential GPU type availability restrictions 
 - Adds a message informing the user that GPU drivers may be installed on machine startup when they request to scale their VM
 - Changes the styling of the hardware scaling form slightly (adding labels and subtitles)

### DIFF
--- a/jupyterlab_gcedetails/src/components/machine_type_select.tsx
+++ b/jupyterlab_gcedetails/src/components/machine_type_select.tsx
@@ -27,7 +27,7 @@ import TextField from '@material-ui/core/TextField';
 import InputAdornment from '@material-ui/core/InputAdornment';
 import Grid from '@material-ui/core/Grid';
 import Paper from '@material-ui/core/Paper';
-import { Option, TEXT_STYLE } from '../data';
+import { Option, TEXT_STYLE, TEXT_LABEL_STYLE } from '../data';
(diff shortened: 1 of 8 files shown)


jupyterlab_gcedetails/src/components/machine_type_select.tsx
```diff
@@ -270,8 +271,12 @@ export class NestedSelect extends React.Component<
 
     return (
       <div className={STYLES.nestedSelect}>
-        {label && <label>{label}</label>}
+        {/* {label && <label>{label}</label>} */}
```
prodonjs, line 274: Can you remove this rather than leave it commented if it's not used?

yaminahq: Ah that's my bad, I forgot to remove it - good catch!

PR #203, opened 2020-08-06
[BigQuery] block dataframe update until query finishes

MRzNone:


MRzNone:
I also refactored widget manager to decouple the logics of editor number

cbwilkes:
I think there are some conflicts here.

PR #204, opened 2020-08-06
[uCAIP] Add documentation for public Python API

shunr:
- Added documentation for and refactored public API
- Added npm script to generate README from docstrings
Dependent on #183


jupyterlab_automl/jupyterlab_automl/__init__.py
```diff
@@ -4,7 +4,7 @@
 from jupyterlab_automl.handlers import handlers
 from jupyterlab_automl.version import VERSION
 from jupyterlab_automl.public import *
-from jupyterlab_automl.service import ModelFramework
+from jupyterlab_automl.types import *
```
timothyrjames, line 7: nit: generally not a great practice to import *

see: https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/

PR #215, opened 2020-08-07
Bugfix/update widget data

l1ghtspeed:


jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -159,4 +164,26 @@ export class HardwareScalingStatus extends React.Component<Props, State> {
...
     );
   }
+
+  private async serverStartup(): Promise<boolean> {
+    let attempt = 0;
+    const numRetries = 20;
+    return new Promise((resolve, reject) => {
+      const interval = setInterval(async () => {
+        try {
+          await this.props.detailsServer.getUtilizationData();
```
l1ghtspeed, line 174: So I'm just currently waiting until this get details request succeeds, not sure if there is a better API to use for this? Thought that since this is the information we want to make sure we had anyways it would make sense to check that this is the request that succeeds.

prodonjs: So I like the approach, but there is a bit of a shortcut you can do to accomplish a loop with a "sleep" in it.
(I'm writing this out in the comment so don't yell at me if it doesn't compile :))

```
for (const i = 0; i < 20; i++) {
   try {
     await this.props.detailsServer.getUtilizationData();
     break;
   } catch (err) (
      if (i === 19) {
        throw err;
      }
      await new Promise(r => setTimeout(r, 5000)); // waits 5s
   )
}
```

PR #219, opened 2020-08-07
[BigQuery] Display dates and times according to user timezone

CynthiaXJia:
Uses [luxon](https://moment.github.io/luxon/index.html), a [successor](https://moment.github.io/luxon/docs/manual/why.html) of [moment](https://momentjs.com/) to detect the user's timezone and apply it. This is done in a new `formatters` file under the `utils` folder.

Updated python tests as well since the date is now passed as an ISO format string rather than a pre-formatted date string.

### DIFF
--- a/jupyterlab_bigquery/src/utils/formatters.ts
+++ b/jupyterlab_bigquery/src/utils/formatters.ts
@@ -1,0 +1,6 @@
+import { DateTime } from 'luxon';
+
+export function formatDate(dateString) {
+  const date = DateTime.fromISO(dateString);
+  return date.toLocaleString(DateTime.DATETIME_MED_WITH_SECONDS);
+}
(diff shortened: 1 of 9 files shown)


jupyterlab_bigquery/src/utils/formatters.ts
```diff
@@ -0,0 +1,6 @@
+import { DateTime } from 'luxon';
+
+export function formatDate(dateString) {
+  const date = DateTime.fromISO(dateString);
+  return date.toLocaleString(DateTime.DATETIME_MED_WITH_SECONDS);
+}
```
CynthiaXJia, line 6: Not sure if `utils/formatters` is the best place to put this - let me know if there's a more appropriate location. :)

cbwilkes: Sounds good to me. We can always refactor later if need be.

PR #220, opened 2020-08-07
[GitSync] Change editor UI view & service iteration procedure

ashleyswang:
- Changed UI scroll view to match previous scroll view on display updates (as opposed to matching the cursor)
- Added service option to add a delay to single sync iteration & ensure iterations don't overlap
- Added state updates for future status indicator display
- Added state listeners for dirty file/conflict signals


jupyterlab_gitsync/src/components/file.tsx
```diff
@@ -5,6 +5,9 @@ export interface Props {
   path: string;
 }
 
+// TO DO: implement file list of conflicting files
```
yunluyl, line 8: // TODO(ashleyswang)

jupyterlab_gitsync/src/components/file.tsx
```diff
@@ -39,4 +42,4 @@ export class File extends React.Component<Props> {
   private _onResolveClick() {
     console.log('file resolved');
```
yunluyl, line 43: no console.log please. this is why reviewers are hesitate to give approval on this, because it's easy to forget.

jupyterlab_gitsync/src/components/setup.tsx
```diff
@@ -12,6 +12,8 @@ export interface Props {
   service: GitSyncService;
 }
 
+// TO DO: Implement Git Repo setup for sync configuration
```
yunluyl, line 15: // TODO(ashleyswang)

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+import { Props } from './panel';
+
+export interface State {
+  label: string;
+  style: ReturnType<typeof style>;
+  options?: any;
+}
+
+// TO DO: Change to non-inheritance implementation of React Component
```
yunluyl, line 17: TODO

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
```diff
@@ -0,0 +1,74 @@
...
+        >
+          <span className={classes('bp3-button-text')}>
+            <span
+              className={classes(
+                'jp-ToolbarButtonComponent-icon',
+                toolbarButtonIconClass,
+                this.state.style
+              )}
+            >
+              {' '}
```
yunluyl, line 63: Please always remember to add TODOs to un-implemented code. Easy to forget, and hard for reviewers to understand what happens here. Is this a mistake? Is this a placeholder? Is this careless? It will end up with all reviewers commenting on this, and waste a lot of resources.

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
```diff
@@ -0,0 +1,74 @@
...
+            </span>
+          </span>
+        </button>
+      </div>
+    );
+  }
+
+  protected _onClick = (): void => {
+    return;
```
yunluyl, line 72: TODO for un-implemented code

ashleyswang: This is meant to be an abstract class where its subclasses override this function, however Typescript does not allow function declarations without definitions. 

This issue will be addressed when in a future PR. I plan on talking to Jason on Monday about re-implementing this class without using inheritance as per React conventions. There is a TODO at the top of the file indicating this change.

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
```diff
@@ -0,0 +1,74 @@
...
+  toolbarButtonClass,
+  toolbarButtonIconClass,
+} from '../style/toolbar';
+
+import { Props } from './panel';
+
+export interface State {
+  label: string;
+  style: ReturnType<typeof style>;
+  options?: any;
```
ylil93, line 14: It looks like you're using this `options` field to keep track of state (running vs not running). Wouldn't it be cleaner to use an enum or boolean instead.

jupyterlab_gitsync/src/components/toolbar_buttons.tsx
```diff
@@ -1,118 +1,61 @@
...
 
     this.props.service.start();
+    this._addListeners();
   }
 
   protected _onClick = (): void => {
-    if (this.props.service.isRunning) {
-      this._stopSync();
+    if (this.props.service.running) {
+      this.props.service.stop();
```
ylil93, line 21: Seems to me like this toggles the auto sync feature, so if that's the case, wouldn't the state also need to be set here?

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
ashleyswang: Since right now the buttons are pretty simple in functionality it may be simpler to do a boolean right now, but since I plan on having the status indicator to be on the toolbar and will most likely be adding more complex buttons that require more state parameters, I decided it might be easier to do just an optional `options` parameter instead of having many optional parameters specific to each individual button. 

For example, the `ControlButton` (currently used to toggle the auto sync feature) needs to keep track of the running state for the button whereas the status indicator will not need to do this, but instead might have a optional `state` parameter indicating which state of the sync process it is in. Since I'm not quite sure if there will need to be more state parameters for specific button, I thought it might be simpler to just use a single `options` parameter encapsulating all of this. 

Let me know what your thoughts are about doing a general `options` parameter vs. having all the optional parameters explicitly stated.

jupyterlab_gitsync/src/components/toolbar_buttons.tsx
ashleyswang: When we call `service.stop()/.start()`, the service will emit a signal indicating whether or not its state has changed. This is done because there are cases in which I believe the sync should not be able to start even if it is currently stopped, such as if it is in a blocked state due to there being unresolved conflicts in a file. 

Instead, we connect the button to an event listener so that if the service state change is successful, then we are allowed to change the state of the button to match the service. By making the button state depend on the service state, we can ensure that they are always synced and that the button always shows the correct state of the service.

ylil93: Ah I see makes sense.

jupyterlab_gitsync/src/components/toolbar_btn_template.tsx
ylil93: I see. I'm actually not too sure what is preferred in javascript/typescript, but generally, it's preferred to have consistent  typing. I that would make your code more robust, but for now, I'm fine with you leaving it as is.

Please do add the description of the intended use from your last post as a code comment in this file for clarification.

ashleyswang: After talking to Jason, I realize that we can actually get rid of this field once I do the refactoring for the ToolbarButton component. I'll add the description for it for now, but this will be fixed in a future PR.

ylil93: SGTM!

PR #224, opened 2020-08-07
[BigQuery] Open query editor from view and model details

CynthiaXJia:
Allows user to open a new full-window editor with a starter query from the view and model details pages.

### DIFF
--- a/jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
+++ b/jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
@@ -86,7 +86,10 @@ export default class TableDetailsTabs extends React.Component<Props, State> {
                     'main',
                     queryId,
                     undefined,
-                    [queryId, `SELECT * FROM \`${this.props.table_id}\``]
+                    [
+                      queryId,
+                      `SELECT * FROM \`${this.props.table_id}\` LIMIT 1000`,
+                    ]
                   );
                 }}
                 startIcon={<Code />}
--- a/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
+++ b/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
@@ -81,6 +81,7 @@ const SQL_EDITOR_OPTIONS: editor.IEditorConstructionOptions = {
   wrappingIndent: 'same',
   wrappingStrategy: 'advanced',
   minimap: { enabled: false },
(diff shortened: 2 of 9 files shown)


jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
```diff
@@ -86,7 +86,10 @@ export default class TableDetailsTabs extends React.Component<Props, State> {
                     'main',
                     queryId,
                     undefined,
-                    [queryId, `SELECT * FROM \`${this.props.table_id}\``]
+                    [
+                      queryId,
+                      `SELECT * FROM \`${this.props.table_id}\` LIMIT 1000`,
+                    ]
```
CynthiaXJia, line 92: I changed this to `LIMIT 1000` to mimic the web UI behavior

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
```diff
@@ -595,7 +596,7 @@ class QueryTextEditor extends React.Component<
     const { iniQuery } = this.props;
 
     // eslint-disable-next-line no-extra-boolean-cast
-    const queryValue = !!iniQuery ? iniQuery : 'SELECT * FROM *';
+    const queryValue = !!iniQuery ? iniQuery : '';
```
CynthiaXJia, line 599: I think an empty editor works nicer than a `SELECT * FROM *` as Polong noted in a feedback comment:

> After installing the extension and clicking on the extension for the first time to see the console in a new tab, the default syntax is:
SELECT * FROM *
>
>It shows an error, which makes me feel like I've done something wrong. It should instead be blank or use a legitimate query

jupyterlab_bigquery/src/components/details_panel/table_details_tabs.tsx
cbwilkes: Thanks this was a friction point that makes this more usable

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
cbwilkes: An empty query sounds good to me

cbwilkes:
Seems like there are some conflicts to resolve

CynthiaXJia:
> Seems like there are some conflicts to resolve

Should be good to go now!

PR #225, opened 2020-08-10
[HWCONFIG]/update opening dialog flow

l1ghtspeed:


jupyterlab_gcedetails/src/details_widget.tsx
```diff
@@ -84,19 +83,14 @@ export class VmDetails extends React.Component<Props, State> {
...
         <span className={classes(STYLES.interactiveHover)}>
           {details ? this.getDisplayedDetails(details) : noDetailsMessage}
         </span>
-        <span
-          className={classes(STYLES.icon, ICON_CLASS, STYLES.interactiveHover)}
-          title="Show form"
-          onClick={() => this.setState({ formDisplayed: true })}
-        ></span>
         {formDisplayed && (
           <HardwareScalingDialog
```
yaminahq, line 95: Nit but maybe we should change the name of the component from _HardwareScalingDialog_ and the state variable from _showForm_ to something more generic since the dialog now shows the hardware configuration details as well as allows you to modify them. Maybe something like _HardwareConfigurationDialog_ and _showDialog_?

l1ghtspeed:
Here's the link for the screen cast, the styling is not finalized, but we can go over that along with the other styling tomorrow during standup.
https://screencast.googleplex.com/cast/NDk1MDk5ODMxMTIzOTY4MHw4NWU4MjM3OC03YQ

PR #229, opened 2020-08-10
[BigQuery] Explore data

MRzNone:
image

Currently, only supporting explore with data studio since there is no good way to do Google Sheet at this moment.


jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_editor_results.tsx
```diff
@@ -2,10 +2,14 @@
...
 import { stylesheet } from 'typestyle';
 import { connect } from 'react-redux';
-import { QueryResult } from '../query_text_editor/query_text_editor';
+import { QueryResult } from './query_text_editor';
 import { QueryId } from '../../../reducers/queryEditorTabSlice';
 import { Header } from '../../shared/header';
 import { BQTable } from '../../shared/bq_table';
+import { Button, Typography } from '@material-ui/core';
+import { ContextMenu } from 'gcp_jupyterlab_shared';
+import DropDown from '../../shared/dropdown';
```
cbwilkes, line 11: Seems like you add this include but you don't use it anywhere. Can the unused includes be removed?

PR #231, opened 2020-08-11
[HardwareConfiguration] Fixing UI bugs

yaminahq:
- An update to the shared/action_bar.tsx file caused the buttons to overlap with the input on the hardware scaling form. Resolved this by creating an action_bar component specifically for gce_details (we were probably going to do this anyways because we need to add some more gce_details specific functionality to the action_bar):
image

- Testing the extension on a real Notebook Instance showed that the input labels incorporated from material-ui on the hardware scaling form were rendering strangely. Switched to using native labels for now - will look more into this if we decide we want to keep the material-ui labels:
image

Hardware scaling form with changes made in this commit:
Screenshot 2020-08-10 at 9 04 07 PM

### DIFF
--- a/jupyterlab_gcedetails/src/components/action_bar.tsx
+++ b/jupyterlab_gcedetails/src/components/action_bar.tsx
@@ -1,0 +1,54 @@
+import * as csstips from 'csstips';
+import * as React from 'react';
+import { stylesheet, classes } from 'typestyle';
+import { css, COLORS } from 'gcp_jupyterlab_shared';
+
+interface Props {
+  primaryLabel: string;
+  secondaryLabel?: string;
+  onPrimaryClick: () => void;
+  onSecondaryClick?: () => void;
+}
+
+const STYLES = stylesheet({
+  actionBar: {
+    padding: '16px',
+    marginTop: '5px',
+    display: 'block',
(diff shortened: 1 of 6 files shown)


jupyterlab_gcedetails/src/components/action_bar.tsx
```diff
@@ -0,0 +1,54 @@
...
+    ...csstips.endJustified,
+  },
+  primaryButton: {
+    backgroundColor: COLORS.blue,
+    color: COLORS.white,
+  },
+});
+
+/** Funtional Component for defining an action bar with buttons. */
+export function ActionBar(props: Props) {
```
prodonjs, line 33: The shared package has an ActionBar component as well. I would prefer that we parameterize it more to suit our purposes if possible, but I'm OK if we do so in a later PR.

It currently does support adding additional buttons, but it may be that we need more fine-grained control of the styling.

yaminahq: Yeah we were originally using this action bar component from the shared package but it's styling was recently updated which caused the rendering issue shown in the first picture. But I can also look into parameterizing the styling for the shared action bar to resolve that issue

PR #232, opened 2020-08-11
[HWCONFIG]/close auth popup error

l1ghtspeed:


jupyterlab_gcedetails/src/components/auth_token_retrieval.tsx
```diff
@@ -23,5 +25,12 @@ export function authTokenRetrieval(): Promise<string> {
       '_authPopup',
       'left=100,top=100,width=400,height=400'
     );
+    timer = setInterval(function() {
```
l1ghtspeed, line 28: So apparently if I try setting a function for authPopup.beforeUnload, it won't work because the child window's domain is different from the parents. I searched it up and people are saying this is the way to go. Tested it and it works.

prodonjs: Great investigation. I'm good with it!

jupyterlab_gcedetails/src/components/auth_token_retrieval.tsx
```diff
@@ -1,6 +1,7 @@
 export function authTokenRetrieval(): Promise<string> {
   // eslint-disable-next-line prefer-const
   let authPopup: any;
```
prodonjs, line 3: Can we add the proper types to these definitions. I believe `authPopup` will be a Window type and `timer` should be a number. You can look at what VSCode shows as the inferred type.

PR #235, opened 2020-08-11
[GitSync] refactor toolbar button component w/o inheritance

ashleyswang:
- refactor the toolbar button components to be without inheritance as per React convention
- renamed `toolbar_btn_template.tsx` to `toolbar_button.tsx`
- no functionality changed


ashleyswang:
@ylil93 
> LGTM'd
> 
> As a side note, to me, it still seems like there's inheritance going on, so I'm wondering what is the main difference after this refactor?

I believe that the difference is we are not explicitly defining the `ControlButton/SyncButton/etc.` as a subclass of the main `ToolbarButton` class, but instead we just call its constructor to do the work for us. That way instead of overriding functions to provide specific functionality, we instead pass in parameters in the `ToolbarButtonProps` to call on functions we define in the `ControlButton` class (just as an example). 

I think that @prodonjs might have a better explanation of this, but I that's the main idea.

ylil93:
> @ylil93
> 
> > LGTM'd
> > As a side note, to me, it still seems like there's inheritance going on, so I'm wondering what is the main difference after this refactor?
> 
> I believe that the difference is we are not explicitly defining the `ControlButton/SyncButton/etc.` as a subclass of the main `ToolbarButton` class, but instead we just call its constructor to do the work for us. That way instead of overriding functions to provide specific functionality, we instead pass in parameters in the `ToolbarButtonProps` to call on functions we define in the `ControlButton` class (just as an example).
> 
> I think that @prodonjs might have a better explanation of this, but I that's the main idea.

Cool thanks for the explanation.

prodonjs:
> > @ylil93
> > > LGTM'd
> > > As a side note, to me, it still seems like there's inheritance going on, so I'm wondering what is the main difference after this refactor?
> > 
> > 
> > I believe that the difference is we are not explicitly defining the `ControlButton/SyncButton/etc.` as a subclass of the main `ToolbarButton` class, but instead we just call its constructor to do the work for us. That way instead of overriding functions to provide specific functionality, we instead pass in parameters in the `ToolbarButtonProps` to call on functions we define in the `ControlButton` class (just as an example).
> > I think that @prodonjs might have a better explanation of this, but I that's the main idea.
> 
> Cool thanks for the explanation.

Yep, React generally favors a [composition over inheritance model](https://reactjs.org/docs/composition-vs-inheritance.html#so-what-about-inheritance).

PR #236, opened 2020-08-11
[BigQuery] Query history improvements

CynthiaXJia:
- Pulls out shared read-only editor (between details panel and expanded query history)
- Batches query history by date on the frontend with local timezone accounted for
- Can open old query in a full-window editor
- Adds indication of what caused an error for failed queries
- Adds extra test for query details with an error

A copy to clipboard button will be added in a future PR.

Successful query:
success

Failed query:
failed

### DIFF
--- a/jupyterlab_bigquery/src/components/query_history/query_history_panel.tsx
+++ b/jupyterlab_bigquery/src/components/query_history/query_history_panel.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
-import { Paper, Collapse, LinearProgress } from '@material-ui/core';
+import { Paper, Collapse, LinearProgress, Icon } from '@material-ui/core';
 import { CheckCircle, Error } from '@material-ui/icons';
 import { stylesheet } from 'typestyle';
+import { DateTime } from 'luxon';
 
 import {
   QueryHistoryService,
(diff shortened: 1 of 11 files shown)


jupyterlab_bigquery/src/components/query_history/query_history_panel.tsx
```diff
@@ -63,63 +114,177 @@ interface Props {
...
+          fontSize="default"
+          htmlColor="rgb(230, 0, 0)"
+          className={localStyles.icon}
+        />
+        {props.errorMsg}
+      </div>
+    </Paper>
+  );
+};
+
```
CynthiaXJia, line 145: This is the box at the top of the expanded query details that shows what error caused the query to fail. We may pull this out as a shared component later since it seems the empty schema and empty table display uses a similar format.

jupyterlab_bigquery/src/components/query_history/query_history_panel.tsx
```diff
@@ -63,63 +114,177 @@ interface Props {
...
+          onClick={() => {
+            const queryId = generateQueryId();
+            WidgetManager.getInstance().launchWidget(
+              QueryEditorTabWidget,
+              'main',
+              queryId,
+              undefined,
+              [queryId, query]
+            );
+          }}
```
CynthiaXJia, line 194: Right now no flags are passed, though as we start supporting legacy SQL it will be important to pass that as an extra argument here.

jupyterlab_bigquery/src/components/query_history/query_history_panel.tsx
```diff
@@ -150,6 +315,20 @@ class QueryHistoryPanel extends React.Component<Props, State> {
...
+      const day = date.toLocaleString(DateTime.DATE_SHORT);
+      if (day in queriesByDate) {
+        queriesByDate[day].push(jobId);
+      } else {
+        queriesByDate[day] = [jobId];
+      }
+    });
+    return queriesByDate;
+  };
+
```
CynthiaXJia, line 331: Groups queries by date

PR #237, opened 2020-08-11
[BigQuery] Change fonts to roboto and pull out shared tables

CynthiaXJia:
Changes all the fonts to roboto by importing them from the shared package. It may eventually make more sense to have an overall material UI theme and use a `ThemeProvider` rather than styling all the shared material UI components, but seeing as we have to set the font within each of the widgets already, it is a little simpler this way.


cbwilkes:
Can you fix the conflict here?

CynthiaXJia:
> Can you fix the conflict here?

Should be good to go!

PR #238, opened 2020-08-11
[uCAIP] Dataset component and model test to material UI

josie-garza:
- Simpler deploy model view that allows for customization or basic default options
- Added refresh while model is deploying
- Added algorithm options and simplified dataset component
- Moved all UI for these components to material UI

image
image


jupyterlab_ucaip/jupyterlab_ucaip/service.py
```diff
@@ -507,6 +519,11 @@ def create_training_pipeline(self, training_pipeline_name, dataset_id,
         "optimizationObjective": objective,
     }
 
+    if objective == "maximize-precision-at-recall":
+      training_task_inputs["optimizationObjectiveRecallValue"] = 0.5
+    if objective == "maximize-recall-at-precision":
```
timothyrjames, line 524: elif?

jupyterlab_ucaip/jupyterlab_ucaip/service.py
```diff
@@ -404,11 +404,23 @@ def get_endpoints(self, model_id):
...
+      built = self._build_endpoint("None", endpoint)
+      endpoints.append(built)
+    return endpoints
+
+  def check_deploying(self, model_name, endpoint_id):
     endpoints = self._endpoint_client.list_endpoints(parent=self._parent).endpoints
-    name = "ucaip-extension/" + model_name
-    filtered = filter(lambda x: name == x.display_name, endpoints)
-    filtered = list(filtered)
+    if endpoint_id != "":
```
timothyrjames, line 417: should this be just:

```Python
if endpoint_id:
```

jupyterlab_ucaip/jupyterlab_ucaip/service.py
```diff
@@ -404,11 +404,23 @@ def get_endpoints(self, model_id):
...
-  def check_deploying(self, model_name):
+  def get_all_endpoints(self):
+    gcp_endpoints = self._endpoint_client.list_endpoints(parent=self._parent).endpoints
+    endpoints = []
+    for endpoint in gcp_endpoints:
+      built = self._build_endpoint("None", endpoint)
+      endpoints.append(built)
+    return endpoints
+
+  def check_deploying(self, model_name, endpoint_id):
```
timothyrjames, line 415: It's not clear to me on its face what this method is doing. check_deploying seems like it would return a boolean, not a list. Can we rename or clarify with comments?

jupyterlab_ucaip/jupyterlab_ucaip/service.py
josie-garza: yup! cool I didn't know you could do that in python

jupyterlab_ucaip/jupyterlab_ucaip/service.py
josie-garza: A rename would be good here you're right

PR #240, opened 2020-08-11
Move StripedRows to shared

CynthiaXJia:
Takes `jupyerlab_bigquery/src/shared/striped_rows` and makes it into a shared component. Would likely be reused by the uCAIP extension which has a similar details panel to BigQuery.

### DIFF
--- a/jupyterlab_bigquery/package.json
+++ b/jupyterlab_bigquery/package.json
@@ -66,7 +66,7 @@
     "@reduxjs/toolkit": "^1.4.0",
     "copy-to-clipboard": "^3.3.1",
     "csstips": "1.2.0",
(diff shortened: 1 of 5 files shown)


CynthiaXJia:
> This looks good. Can you rebase this branch on top of master to get all of the upstream changes?

Should be rebased now! Also, it seems like the shared package version was updated to 1.0.6 in [this PR](https://github.com/GoogleCloudPlatform/jupyter-extensions/commit/a3b0eb8cd165bfe19817c9a7da2b113797b97619), so should I bump this version to 1.0.7, or leave it as 1.0.6?

prodonjs:
Closing this as it's quite dated at this point. We can consider reopening if we wish to move striped rows for other extensions later.

PR #241, opened 2020-08-12
[HardwareConfiguration] Adding error handling in rescaling dialog

yaminahq:
Adds error handling for when Notebook API calls fail:
 - If  fails, the instance will be started and the user will be shown the error and the current state of their hardware configuration:
  - https://screencast.googleplex.com/cast/NjMzNzE5NDM3NTUxMjA2NHxiZjc3OGMwZC0wZg
 - If starting the machine fails, the user will be stuck on a dialog asking them to start their machine on the cloud console: 
  - https://screencast.googleplex.com/cast/NTEzMjgxNjg5MzM0NTc5MnxhNGRlMDBiOC1hYw

We can decide on the styling we want for the dialog contents and update it in a different PR

### DIFF
--- a/jupyterlab_gcedetails/src/components/hardware_configuration_dialog.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_configuration_dialog.tsx
@@ -18,7 +18,7 @@ import { Dialog } from '@material-ui/core';
 import * as React from 'react';
 import { HardwareScalingForm } from './hardware_scaling_form';
 import { HardwareScalingStatus } from './hardware_scaling_status';
-import { NotebooksService } from '../service/notebooks_service';
+import { NotebooksService, Instance } from '../service/notebooks_service';
 import {
   HardwareConfiguration,
   Details,
@@ -27,12 +27,27 @@ import {
 import { ConfirmationPage } from './confirmation_page';
 import { ServerWrapper } from './server_wrapper';
 import { DetailsDialogBody } from './details_dialog_body';
+import { ErrorPage } from './error_page';
 
 enum View {
   DETAILS,
   FORM,
   CONFIRMATION,
   STATUS,
+  ERROR,
+}
+
+export enum ErrorType {
+  STOP = 'Stop',
+  RESHAPING = 'Reshape',
+  START = 'Start',
+  REFRESH = 'Refresh',
+}
+
+export interface ConfigurationError {
+  errorType: ErrorType;
+  errorValue: any;
+  instanceDetails?: Instance;
 }
 
 interface Props {
@@ -48,15 +63,16 @@ interface Props {
 interface State {
   view: View;
   hardwareConfiguration: HardwareConfiguration;
+  error: ConfigurationError;
 }
 
-/** Funtional Component for a common dialog interface with cancel and submit buttons. */
 export class HardwareConfigurationDialog extends React.Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = {
       view: View.DETAILS,
       hardwareConfiguration: null,
+      error: null,
     };
   }
 
@@ -64,6 +80,20 @@ export class HardwareConfigurationDialog extends React.Component<Props, State> {
     const { open } = this.props;
 
     return <Dialog open={open}>{this.getDisplay()}</Dialog>;
+  }
+
+  private onError(
+    errorValue,
+    errorType: ErrorType,
+    instanceDetails?: Instance
+  ) {
+    this.setState({
+      error: {
+        errorValue,
+        errorType,
+        instanceDetails,
+      },
+    });
   }
 
   private getDisplay() {
@@ -75,7 +105,7 @@ export class HardwareConfigurationDialog extends React.Component<Props, State> {
       detailsServer,
       receivedError,
     } = this.props;
-    const { view, hardwareConfiguration } = this.state;
+    const { view, hardwareConfiguration, error } = this.state;
 
     switch (view) {
       case View.DETAILS:
@@ -104,6 +134,7 @@ export class HardwareConfigurationDialog extends React.Component<Props, State> {
             details={details}
           />
         );
+
       case View.CONFIRMATION:
         return (
           <ConfirmationPage
@@ -119,6 +150,7 @@ export class HardwareConfigurationDialog extends React.Component<Props, State> {
             }}
           />
         );
+
       case View.STATUS:
         return (
           <HardwareScalingStatus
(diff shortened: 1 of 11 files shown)


jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -105,46 +112,101 @@ export class HardwareScalingStatus extends React.Component<Props, State> {
     window.addEventListener('beforeunload', this.preventPageClose);
   }
 
-  async componentDidUpdate() {
-    const { hardwareConfiguration, notebookService, onCompletion } = this.props;
-    const { machineType, gpuType, gpuCount, attachGpu } = hardwareConfiguration;
+  async componentDidUpdate(prevProps, prevState) {
```
prodonjs, line 115: This method has gotten pretty large and it's starting to feel like maybe we need to refactor to avoid conflating the two concerns of managing the state transitions from one operation to the next with the UX presentation of the current state. We can discuss in our standup.

jupyterlab_gcedetails/src/components/hardware_configuration_dialog.tsx
```diff
@@ -127,8 +159,19 @@ export class HardwareConfigurationDialog extends React.Component<Props, State> {
             notebookService={notebookService}
             onCompletion={onCompletion}
             detailsServer={detailsServer}
+            onError={(err, errorType, instanceDetails?) =>
```
prodonjs, line 162: I'm a bit confused why there are two callbacks passed in that both seem to pertain to error handling. Could the `onError` callback serve both purposes?

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -105,46 +112,101 @@ export class HardwareScalingStatus extends React.Component<Props, State> {
...
-    try {
-      switch (this.state.status) {
-        case Status['Stopping Instance']:
-          await notebookService.stop();
-          this.setState({ status: Status['Reshaping Instance'] });
+    if (prevState.status !== status) {
+      switch (status) {
+        case Status['Stopping Instance']: {
+          try {
+            const stopResult = (await notebookService.stop()) as Instance;
```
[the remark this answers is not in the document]
yaminahq: I put it in just to make it clear that the result of the API call is used as an Instance type object in the rest of the code but its not needed - I can remove it.

jupyterlab_gcedetails/src/components/hardware_configuration_dialog.tsx
yaminahq: The onError callback stores the error that was experienced in the state of the hardware scaling dialog. The showPage callback changes the view of the hardware scaling dialog to be the error page rather than the scaling status page. I have these separated because specifically if an error occurs while setting the machineType/accelerator we want to save the error in the state variable so we can pass it to the error page later but we don't want to show the error page until the start operation has been attempted. I thought it was more modular this way but I can also wrap these two functionalities into one callback that checks what type of error was received and doesn't change the hardware scaling dialog view if the error occurred while setting the machineType/accelerator to make it more concise if that's the better approach?

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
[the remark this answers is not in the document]
prodonjs: Yeah, in general if the type can be inferred from the method's return type, it's not common to do an explicit cast or explicit type annotation on the declaration.

jupyterlab_gcedetails/src/components/hardware_configuration_dialog.tsx
prodonjs: I think this goes along with other comment about the HardwareScalingStatus component being responsible for both the operation and its display. The strange part for me is that the child component is bubbling up a status with the `onError` callback, but also being responsible for knowing when to change its parents state by calling `showErrorPage`. It seems like the parent component should be responsible for knowing when to transition states like it does for the other dialog states.

yaminahq: Maybe it would fair better to move some of the logic in the scaling status component into the dialog component so the scaling status page is just updating the view based on the status of the dialog? We can talk more about this in our meeting

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
yaminahq: I've resolved this for the time being by handling displaying the error page in the hardware scaling status component rather than the dialog. I've also cleaned up the code a bit by using methods to make it clear what is going on in each case of this method. In a future PR I think we should consider moving towards using a state management system like redux as we discussed in our meeting to separate state management and UX concerns.

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -215,15 +262,21 @@ export class HardwareScalingStatus extends React.Component<Props, State> {
   }
 
   render() {
-    const { status } = this.state;
-    const progressValue = (status / 4) * 100;
+    const { status, error, instanceDetails } = this.state;
+    const progressValue = (status / 6) * 100;
```
prodonjs, line 266: This is actually not a constant right? If a user isn't modifying the machine type or the GPU, this might be 1 less step? Or is that a subsequent change?

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
```diff
@@ -215,15 +262,21 @@ export class HardwareScalingStatus extends React.Component<Props, State> {
...
+        onDialogClose={onDialogClose}
+        error={error}
+        instanceDetails={instanceDetails}
+      />
+    ) : (
       <div className={flexContainer}>
         <h3 className={heading}>{Status[status]}</h3>
         <p className={paragraph}>{statusInfo[status]}</p>
-        {status === 4 || status === 5 ? (
+        {status === 6 ? (
```
prodonjs, line 279: Can we refer to the Status by it's Enum value (`Status['Complete']` I believe)? Otherwise, it sort of defeats the purpose of using enums as named constants.

jupyterlab_gcedetails/src/components/hardware_scaling_status.tsx
yaminahq: Yeah this is not necessarily a constant if a user doesn't attach a GPU then that step will be skipped. But because the number corresponding to the Status['Complete'] is 6, the animation always works out with the complete step showing the progress bar as full. For this reason I didn't compute this value specifically for each case. But i can update it to include the case in which the attachGpu step is skipped.

prodonjs: No need to add that now since we expect additional changes as part of the refactoring.

PR #244, opened 2020-08-12
[uCAIP] Autogen code cells and refactor context

shunr:
- Add CodeGenService to easily create Notebook code cells from any component
- Refactored widgets and components to use React's context API instead of passing context down through props
- Add code cell buttons to code components


jupyterlab_ucaip/src/components/base_widget.tsx
```diff
@@ -0,0 +1,34 @@
...
+    super();
+  }
+
+  render() {
+    return (
+      <ThemeProvider theme={theme}>
+        <AppContext.Provider value={this.context}>
+          {this.body()}
+        </AppContext.Provider>
+      </ThemeProvider>
```
josie-garza, line 29: This is also a really good idea in terms of changing the mui theme https://material-ui.com/customization/palette/#dark-mode

jupyterlab_ucaip/src/service/code_gen.ts
```diff
@@ -0,0 +1,49 @@
...
+        NotebookActions.insertBelow(notebook);
+      }
+      notebook.activeCell.model.value.text = code;
+    }
+  }
+
+  static importDatasetCode(datasetId: string) {
+    return `from jupyterlab_ucaip import import_dataset
+df = import_dataset('${datasetId}')`;
+  }
```
josie-garza, line 33: should we move all generated code strings to this service?

PR #247, opened 2020-08-12
Comment context for non-notebook files

mkalil:
working on display comment context for notebook files in a future PR

### DIFF
--- a/jupyterlab_comments/jupyterlab_comments/git_commands.py
+++ b/jupyterlab_comments/jupyterlab_comments/git_commands.py
@@ -125,8 +125,8 @@ class Git(Configurable):
 			review_comments = review.get("comments", [])
 			comments_on_file = []
 			for item in review_comments:
-				location = item["comment"]["location"]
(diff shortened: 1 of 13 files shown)


jupyterlab_comments/jupyterlab_comments/git_commands.py
```diff
@@ -125,8 +125,8 @@ def get_all_code_review_comments(self, file_path_from_repo_root, git_root_dir):
 			review_comments = review.get("comments", [])
 			comments_on_file = []
 			for item in review_comments:
-				location = item["comment"]["location"]
-				if location.get("path", "") == file_path_from_repo_root:
+				location = item["comment"].get("location", "")
```
ojarjur, line 128: Make the default here `{}` instead of `""`. That way, you can chain the `get` methods. You might also want to user `get` for indexing into `item` as well.

e.g. `path = item.get("comment", {}).get("location", {}).get("path", "")`

PR #250, opened 2020-08-13
[HardwareConfig] Style updates

l1ghtspeed:
Just to look at diffs


yaminahq:
I'm working on fixing the missing packages issue right now but correct me if I'm wrong I don't think I can remove the package-lock.json file since it was part of @l1ghtspeed 's commit

shared/client/components/progress.tsx~merged
```diff
@@ -0,0 +1,40 @@
+/**
```
prodonjs, line 1: It looks like there are merge conflicts here that need to be reviewed.

yaminahq:
See https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/254

PR #260, opened 2020-08-18
[BigQuery] Parallel format query.  Improved query speed.  UI still lags a bit due to large data

MRzNone:


MRzNone:
Prevent big query from freezing the UI for too long.  Will need to follow up on reduce UI freezing

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
```diff
@@ -261,7 +261,7 @@ class QueryTextEditor extends React.Component<
           this.setState({ queryState: QueryStates.READY });
         }
       },
-      2000
+      300000
```
cbwilkes, line 264: We should put these values in constants. Like:

```
const QUERY_BATCH_SIZE = 300000;
```

jupyterlab_bigquery/jupyterlab_bigquery/pagedAPI_handler/pagedQuery_handler.py
```diff
@@ -101,8 +107,13 @@ def query(self, request_body, page_size):
     schema_fields = format_preview_fields(en.schema)
 
     for page in en.pages:
+      if page.num_items > 1e5:
```
cbwilkes, line 110: 1e5 is also seems like a "magic" number can you make it const also?

PR #262, opened 2020-08-19
[BigQuery] Dropdown for training runs, training options

CynthiaXJia:
Adds a dropdown for training runs if there are multiple, and makes training options visible.

Also refactors the shared header to account for the dropdown which is located there.

trainingruns

### DIFF
--- a/jupyterlab_bigquery/src/components/details_panel/service/list_model_details.ts
+++ b/jupyterlab_bigquery/src/components/details_panel/service/list_model_details.ts
@@ -13,6 +13,7 @@ export interface ModelDetailsObject {
   model_type: string;
   schema_labels: ModelSchema[];
   feature_columns: ModelSchema[];
+  training_runs: string[];
 }
 
 export interface ModelSchema {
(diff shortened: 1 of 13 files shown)


jupyterlab_bigquery/src/components/details_panel/service/list_model_details.ts
```diff
@@ -24,6 +25,35 @@ export interface ModelDetails {
...
+  learn_rate?: number; // float
+  learn_rate_strategy?: string; // model enum
+  loss_type?: string; // model enum
+  max_iterations?: number; // int
+  min_relative_progress?: number; // DoubleValue
+  model_uri?: string;
+  num_clusters?: number; // int
+  optimization_strategy?: string; // model enum
+  warm_start?: string; // BoolValue
+}
```
CynthiaXJia, line 51: out of all these options, I'm not sure what format `input_label_columns` and `label_class_weights` will come as, but I've labeled all of these particular types with more specific annotations just so its a little more clear where they come from.

cbwilkes: I searched internally. I think this can probably be either a string and array of strings

PR #267, opened 2020-08-20
[HWCONFIG] HardwareScalingStatus Tests

l1ghtspeed:
Also added in @prodonjs 's suggested implementation for the testing the auth token


jupyterlab_gcedetails/src/components/hardware_scaling_status.spec.tsx
```diff
@@ -0,0 +1,218 @@
...
+        onDialogClose={mockOnDialogClose}
+        hardwareConfiguration={hardwareConfig}
+        authTokenRetrieval={mockAuthTokenRetrieval}
+      />
+    );
+    await mockAuthTokenRetrieval();
+    await mockStopInstance();
+    await mockStartInstance();
+    await mockGetUtilizationData();
+    setTimeout(() => {
```
l1ghtspeed, line 176: not sure if there is a better way to do this but it would register the expect too fast before the state actually changed, hence the timeout

jupyterlab_gcedetails/src/components/hardware_scaling_status.spec.tsx
```diff
@@ -0,0 +1,218 @@
...
+      expect(hardwareScalingStatus).toMatchSnapshot('Operation error');
+      hardwareScalingStatus
+        .find(Button)
+        .first()
+        .simulate('click');
+      expect(mockOnDialogClose).toBeCalled();
+    }, 1000);
+  });
+
+  it('Renders with restart error', async () => {
```
l1ghtspeed, line 187: So I tried to go for a testing approach which doesn't cover every single error situation, but it covers all the ones where the UI would be different. Let me know if it should be exhaustive though

jupyterlab_gcedetails/src/components/hardware_scaling_status.spec.tsx
```diff
@@ -0,0 +1,218 @@
...
+      hardwareScalingStatus
+        .find(Button)
+        .first()
+        .simulate('click');
+      expect(mockOnDialogClose).toBeCalled();
+    }, 1000);
+  });
+
+  it('Renders with auth error', async () => {
+    mockAuthTokenRetrieval.mockImplementation(() => {
```
prodonjs, line 107: I think it's more customary and simpler to use mockRejectedvalue or mockReturnValue with a rejected promise.

https://jestjs.io/docs/en/mock-function-api#mockfnmockrejectedvaluevalue

jupyterlab_gcedetails/src/components/hardware_scaling_status.spec.tsx
prodonjs: Generally you want to avoid setting timeouts in your tests because they slow them down considerably. Let's talk about this one in our standup.

jupyterlab_gcedetails/src/components/hardware_scaling_status.spec.tsx
prodonjs: I think this is sufficient as each error is handled similarly regardless of where it occurs.

PR #269, opened 2020-08-20
[BigQuery] speed up query result transfer

MRzNone:


MRzNone:
Significantly reduces frozen UI due to big query.  Still a little lag (< 1s).  Will investigate more.

jupyterlab_bigquery/src/utils/BigDataManager.ts
```diff
@@ -0,0 +1,34 @@
+class BigDataManager {
```
cbwilkes, line 1: I don't know if I like the name BigDataManager. How about `QueryResultsManager` or just `DataManager` if it is generic.

MRzNone: I was intended to be a generic class, but I found different info need different format.  I renamed it to `QueryResultsManager`.

PR #270, opened 2020-08-20
[BigQuery] re-use client in query history

MRzNone:


MRzNone:
It speeds up start-up time for query history by approximately 5 seconds.

PR #274, opened 2020-08-21
[GitSync] nbmerge client-side implementation

ashleyswang:
* remove server-side nbmerge files
* implement client-side nbmerge without nbmerge API


jupyterlab_gitsync/src/service/file.ts
```diff
@@ -91,7 +101,7 @@ export class File implements IFile {
       const contents = await fs.get(this.path);
       this.resolver.addVersion(contents.content, 'remote');
     } catch (error) {
-      console.log(error);
+      console.warn(error);
```
yunluyl, line 104: UI error needs to be exposed to users. Console.warn is not a good way to expose, since end users normally don't open developer console.

jupyterlab_gitsync/src/service/file.ts
```diff
@@ -53,7 +57,9 @@ export class File implements IFile {
 
   async save() {
     try{
-      await this.context.save();
+      const text = this.doc.getValue();
+      await this._saveFile(text);
+      this.resolver.addVersion(text, 'base');
     } catch (error) {
       console.warn(error);
```
yunluyl, line 64: Same here. Try to avoid using console output except for debugging.

jupyterlab_gitsync/src/service/notebook_file.ts
```diff
@@ -1,31 +1,40 @@
...
+  content: Notebook;
   context: DocumentRegistry.Context;
-  resolver: IResolver;
+  resolver: NotebookResolver;
+  view: {
+    left: number,
+    top: number
+  }
+  cursor: {
+    index: number,
```
yunluyl, line 21: my curiosity, what is a cursor index?

jupyterlab_gitsync/src/service/file.ts
```diff
@@ -23,6 +23,10 @@ export class File implements IFile {
     right: number,
     bottom: number 
   }
+  cursor: {
+    line: number,
+    ch: number
```
yunluyl, line 28: you used line and column in notebook file. Let's use consistent names.

jupyterlab_gitsync/src/service/notebook_file.ts
```diff
@@ -41,55 +50,100 @@ export class NotebookFile implements IFile {
...
+  }
+
   async save() {
     try{
-      await this.context.save();
+      const content = (this.content.model as NotebookModel).toJSON();
+      await this._saveFile(content);
+      this.resolver.addVersion(content, 'base');
     } catch (error) {
       console.warn(error);
```
yunluyl, line 67: find a way to notify user

jupyterlab_gitsync/src/service/notebook_file.ts
```diff
@@ -41,55 +50,100 @@ export class NotebookFile implements IFile {
...
     this._getLocalVersion();
-    this._getRemoteVersion();
+    this._getEditorView();
+    const merged = await this.resolver.mergeVersions();
+    if (merged) { await this._displayText(merged); }
+  }
+
+  private async _displayText(merged) {
+    await this._saveFile(merged)
     await this.context.revert();
```
yunluyl, line 81: what does this context revert do?

jupyterlab_gitsync/src/service/notebook_file.ts
```diff
@@ -41,55 +50,100 @@ export class NotebookFile implements IFile {
...
 
   private async _getRemoteVersion() {
-    // TO DO (ashleyswang): change to get raw string of notebook file
-    const contents = await fs.get(this.path, {content: false});
-    console.log(contents.content);
+    try{
+      const contents = await fs.get(this.path);
+      this.resolver.addVersion(contents.content, 'remote');
+    } catch (error) {
+      console.warn(error);
```
yunluyl, line 107: you probably can build a centralized place to display all the error messages. maybe together with the status indicator? so when the status is red, users get some error messages.

jupyterlab_gitsync/src/service/notebook_file.ts
ashleyswang: `index` is the cell index in which the cursor lies. Since each cell has it's own editor, we need to get the index of the cell as well as it's line and character (column) position.

jupyterlab_gitsync/src/service/file.ts
ashleyswang: I use `ch` and `column` because for one I use the `CodeMirror` API directly whereas the other I use the `CodeMirrorEditor` wrapper. I can see about only using the CodeMirror API in the notebook files in a separate PR.

jupyterlab_gitsync/src/service/notebook_file.ts
ashleyswang: context revert displays the disk contents to the user. This is necessary so that when we save, we do not get a pop-up stating that the disk contents have changed since the last time we opened the file (something to do with the last modified time).

jupyterlab_gitsync/src/service/file.ts
ashleyswang: I'm thinking of displaying the errors somewhere on the UI (maybe clicking the status indicator will show a panel with a log of all service operations, ie. save/sync/etc.), but this will most likely be included in a separate PR.

jupyterlab_gitsync/src/service/file.ts
ashleyswang: I'm thinking of displaying the errors somewhere on the UI (maybe clicking the status indicator will show a panel with a log of all service operations, ie. save/sync/etc.), but this will most likely be included in a separate PR.

jupyterlab_gitsync/src/service/notebook_file.ts
ashleyswang: I'm thinking of displaying the errors somewhere on the UI (maybe clicking the status indicator will show a panel with a log of all service operations, ie. save/sync/etc.), but this will most likely be included in a separate PR.

jupyterlab_gitsync/src/service/notebook_file.ts
ashleyswang: I'm thinking of displaying the errors somewhere on the UI (maybe clicking the status indicator will show a panel with a log of all service operations, ie. save/sync/etc.), but this will most likely be included in a separate PR.

PR #279, opened 2020-08-24
Add package-lock.json for jupyterlab_caip_optimizer

prodonjs:
Postpush Actions are failing (https://github.com/GoogleCloudPlatform/jupyter-extensions/actions) and the source is tests from the Optimizer extension. Adding this package-lock.json file should ensure the proper dependencies are installed to build and run Optimizer tests.


prodonjs:
This still doesn't appear to fix the problem. Closing the PR and getting back to the drawing board.

PR #282, opened 2020-08-24
Repo health updates

prodonjs:
- Adds [jupyterlab_gitsync](https://github.com/GoogleCloudPlatform/jupyter-extensions/tree/master/jupyterlab_gitsync) to build scripts and formats existing files
  according to repo-wide lint rules
- Fixes broken Optimizer test
- Updates Jest to version 26
- Updates all `package-lock.json files` to ensure consistent builds in new environments


prodonjs:
@ashleyswang - This is a large PR which touches most of the files in your extension. All I did was check in the linting fixes that are applied to all packages. Then, I made sure that we are actually building and testing `jupyterlab_gitsync` with each presubmit and postsubmit run.

PR #284, opened 2020-08-24
[HardwareConfiguration] Fetching gpus from client

yaminahq:
Removing the code to fetch the list of available GPUs using gcloud compute in the server and instead fetching them in the client using the Compute API.


prodonjs:
@yaminahq Please rebase this to pull in the formatting changes to the handlers files.

jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
```diff
@@ -30,11 +30,9 @@
     'http://metadata.google.internal') + '/computeMetadata/v1/?recursive=true'
 METADATA_HEADER = {'Metadata-Flavor': 'Google'}
 MACHINE_TYPE_CMD = 'gcloud compute machine-types describe'
```
prodonjs, line 32: Not in scope for this change, but do we still need to use the machine types describe call?

PR #286, opened 2020-08-24
[HardwareConfiguration] Displaying instance details in reshape complete message

yaminahq:
Adding new instance details to the message displayed in the dialog once reshaping completes successfully. These details are already displayed in the error page so I'm moving the function that formats them out into it's own component to be used by both the error page and the completion message.

Screenshot 2020-08-24 at 6 28 58 PM

### DIFF
--- a/jupyterlab_gcedetails/src/components/instance_details_message.tsx
+++ b/jupyterlab_gcedetails/src/components/instance_details_message.tsx
@@ -1,0 +1,53 @@
+/**
+ * Copyright 2020 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


jupyterlab_gcedetails/src/components/instance_details_message.tsx
```diff
@@ -0,0 +1,53 @@
...
+  MachineTypeConfiguration,
+  getMachineTypeText,
+} from '../data/machine_types';
+import { STYLES } from '../data/styles';
+import { getGpuTypeText } from '../data/accelerator_types';
+
+export function displayInstance(
+  instance: Instance,
+  machineTypes: MachineTypeConfiguration[],
+  title: string
```
prodonjs, line 30: Seems like this is the same value in both places. Could you provide a default?

prodonjs: Resolved in discussion

PR #291, opened 2020-08-26
[BigQuery] Common info card

CynthiaXJia:
Pulls out the error box from query history and makes it a common component, `InfoCard`, reused for empty tables and partitioned tables.

`InfoCard` takes in a color, icon, message, and optional button, and displays the message along with matching color band and icon.

dismissablepartitioned
emptytable

### DIFF
--- a/jupyterlab_bigquery/src/components/shared/info_card.tsx
+++ b/jupyterlab_bigquery/src/components/shared/info_card.tsx
@@ -24,13 +24,15 @@ const localStyles = stylesheet({
   },
 });
 
+// TODO: figure out type so only material ui icons accepted,
+// and make it work with using .type
+type MaterialUIIcon = any;
+
 interface InfoCardProps {
   message: React.ReactNode;
   button?: React.ReactNode;
   color: string;
-  // TODO: figure out type so only material ui icons accepted,
-  // and make it work with using .type
-  icon: any;
+  icon: MaterialUIIcon;
 }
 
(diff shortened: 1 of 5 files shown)


jupyterlab_bigquery/src/components/shared/info_card.tsx
```diff
@@ -0,0 +1,58 @@
...
+  },
+});
+
+interface InfoCardProps {
+  message: React.ReactNode;
+  button?: React.ReactNode;
+  color: string;
+  // TODO: figure out type so only material ui icons accepted,
+  // and make it work with using .type
+  icon: any;
```
CynthiaXJia, line 33: I couldn't figure out how to type the passed in icon so that I would be able to use `icon.type`...also ideally I would type this so that only icon type components can be passed in, but I'm not sure if that's possible.

cbwilkes: Maybe you could make a placeholder type until then. At least other devs would know the purpose:

```
// TODO: figure out type so only material ui icons accepted,
// and make it work with using .type
type MaterialUIIcon = any;

interface InfoCardProps {
  message: React.ReactNode;
  button?: React.ReactNode;
  color: string;
  icon: MaterialUIIcon;
}
```

CynthiaXJia: Sounds good

PR #293, opened 2020-08-26
[HardwareConfiguration] Adding cost estimation

yaminahq:
Adding PriceService which creates a map of  Compute Resource SKU's to the resource cost and uses it to calculate an estimation of the monthly cost of a particular instance. The estimated monthly cost for an instance is then displayed in the hardware scaling form under a 'Pricing Estimation' section:
https://screencast.googleplex.com/cast/NjMwMzc0NjA0MjYyNjA0OHwyOWE0OGEzOS1mOQ

Will add tests in next PR.

### DIFF
--- a/jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
+++ b/jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
@@ -130,6 +130,18 @@ async def get_gpu_details():
 
   return details
 
+async def get_pricelist():
+  """Retrieves JSON-formatted pricelist data from cloud pricing calculator."""
+  client = AsyncHTTPClient()
+  pricelist = {}
+  try:
+    app_log.debug('Retrieving pricelist')
+    response = await client.fetch("https://cloudpricingcalculator.appspot.com/static/data/pricelist.json")
+    pricelist = json.loads(response.body)
+  except HTTPClientError as e:
+    app_log.error('Unable to retrieve pricelist %s: %s', e.code, e.message)
+  return pricelist
+
 class VmDetailsHandler(APIHandler):
   """Handler to obtain details from the metadata server and local system."""
 
@@ -152,3 +164,16 @@ class VmDetailsHandler(APIHandler):
     self.details[UTILIZATION] = get_resource_utilization()
     self.details[GPU] = await get_gpu_details()
     self.finish(self.details)
+
+class PricelistHandler(APIHandler):
+  """Returns pricing information for instance price estimation."""
+
+  async def get(self):
+    try:
+      pricelist = await get_pricelist()
+      response = pricelist.get('gcp_price_list', {})
+      self.finish(response)
+    except:
+      msg = 'Unexpected error obtaining pricelist'
+      app_log.exception(msg)
+      self.set_status(500, msg)
--- a/jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
+++ b/jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
@@ -171,8 +171,7 @@ class PricelistHandler(APIHandler):
   async def get(self):
     try:
       pricelist = await get_pricelist()
-      response = pricelist.get('gcp_price_list', {})
-      self.finish(response)
+      self.finish(pricelist)
     except:
       msg = 'Unexpected error obtaining pricelist'
       app_log.exception(msg)
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
@@ -209,15 +209,15 @@ export class HardwareScalingForm extends React.Component<Props, State> {
     const zone = details.instance.zone.split('/').pop();
     const currentPrice = priceService.getPrice(zone, this.oldConfiguration);
     const newPrice = priceService.getPrice(zone, configuration);
-    const priceDifference = (newPrice - currentPrice).toFixed(2);
+    const priceDifference = (newPrice - currentPrice);
 
     return (
       <div>
-        <span className={STYLES.subheading}>Pricing estimation:</span>
+        <span className={STYLES.subheading}>Pricing Estimation:</span>
         <div className={STYLES.paragraph}>
           {`Your updated instance will cost an estimated
-          $${priceDifference} monthly, an estimated difference of 
-          $${priceDifference} from your current instance.`}
+          $${newPrice.toFixed(2)} monthly, an estimated difference of 
+          $${priceDifference.toFixed(2)} from your current instance.`}
         </div>
       </div>
     );
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
@@ -32,6 +32,7 @@ import {
   Details,
   detailsToHardwareConfiguration,
   isEqualHardwareConfiguration,
+  extractLast,
 } from '../data/data';
 import {
   ACCELERATOR_COUNTS_1_2_4_8,
@@ -60,6 +61,7 @@ interface Props {
 interface State {
   configuration: HardwareConfiguration;
   gpuCountOptions: Option[];
+  newConfigurationPrice: number | undefined;
 }
 
 export const FORM_STYLES = stylesheet({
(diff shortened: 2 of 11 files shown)


jupyterlab_gcedetails/jupyterlab_gcedetails/handlers.py
```diff
@@ -152,3 +164,16 @@ async def get(self):
...
     self.details[GPU] = await get_gpu_details()
     self.finish(self.details)
+
+class PricelistHandler(APIHandler):
+  """Returns pricing information for instance price estimation."""
+
+  async def get(self):
+    try:
+      pricelist = await get_pricelist()
+      response = pricelist.get('gcp_price_list', {})
```
prodonjs, line 174: I would probably extract the `gcp_price_list` key from the returned JSON object in the get_pricelist function just to keep the logic in the handler as simple as possible and bury the implementation specifics for the endpoint in the accessor function.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -200,11 +202,37 @@ export class HardwareScalingForm extends React.Component<Props, State> {
     this.props.onSubmit(configuration);
   }
 
+  private displayPricingEstimation() {
+    const { details, priceService } = this.props;
+    const { configuration } = this.state;
+
+    const zone = details.instance.zone.split('/').pop();
```
prodonjs, line 209: Just out of curiosity, I know we use this `split('/').pop()` convention in quite a few places, do you think it would make sense to create a helper function for it?

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -200,11 +202,37 @@ export class HardwareScalingForm extends React.Component<Props, State> {
...
+    const currentPrice = priceService.getPrice(zone, this.oldConfiguration);
+    const newPrice = priceService.getPrice(zone, configuration);
+    const priceDifference = (newPrice - currentPrice);
+
+    return (
+      <div>
+        <span className={STYLES.subheading}>Pricing Estimation:</span>
+        <div className={STYLES.paragraph}>
+          {`Your updated instance will cost an estimated
+          $${newPrice.toFixed(2)} monthly, an estimated difference of
```
prodonjs, line 219: At least to me, difference makes me think that we're always subtracting from the current price. Can we make it say an "increase/decrease of" and then show the absolute value of the difference?

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,219 @@
...
+import { ServerConnection } from '@jupyterlab/services';
+import { Option, HardwareConfiguration } from '../data/data';
+
+/** The monthly price catalog for region and sku */
+type Catalog = Map<string, Map<string, number>>;
+
+/** The average number of hours in a month, i.e., 365 / 12 * 24 */
+const HOURS_PER_MONTH = 730;
+const NO_SUSTAINED_DISCOUNT_PREFIXES = ['e2-'];
+const REGIONS: Option[] = [
```
prodonjs, line 26: I'd suggest moving these out to the data constants file since they may eventually prove useful outside of prices.

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,219 @@
...
+    value: 'asia-southeast1',
+    text: 'asia-southeast1 (Singapore)',
+  },
+];
+
+export class PriceService {
+  private catalog: Catalog;
+  private _receivedError: boolean;
+
+  async getPriceList(): Promise<void> {
```
prodonjs, line 113: This method is a bit of misnomer since it really doesn't return anything like the prefix `get` would entail. I'd suggest renaming it to something like loadPrices, or perhaps consider implementing the lazy-loading strategy from some of the other async methods where the first call retrieves the content from the remote API and the subsequent calls return the same promise.

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,219 @@
...
+  private getRegionFromZone(zone: string): string {
+    return zone.split('-', 2).join('-');
+  }
+
+  /**
+   * Returns estimated total monthly price for a given instance to be created in a
+   * certain region.
+   *
+   * Logics adapted from
+   * //depot/google3/java/com/google/developers/console/web/common/compute/pricing/pricing_service.ts
```
prodonjs, line 190: Don't include any references to google3 paths.

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  getPrice(zone: string, config: HardwareConfiguration): number {
```
prodonjs, line 192: The signature for this should be number|undefined.

So the more I think about this, the more I think that this should be the only public method on this service. In Angular, we have the concept of route resolvers that load all of the data needed for a particular page. Here, we are probably OK with waiting until a user explicitly wants to make a hardware config change before fetching the data.

I'm thinking that the actual price retrieval logic is private, and that when this method is called, we use a private Promise data member that has the Catalog. If the Catalog is empty, this method will return undefined as it currently does, otherwise the same logic will apply.

In this way, we won't need to preload the data or check for an error state. It'll be a best attempt to determine pricing, but if we can't load the data for some reason or the configuration doesn't exist, we'll show nothing.

jupyterlab_gcedetails/src/service/price_service.ts
yaminahq: Great, I'll update the name but I believe the way I have implemented does use cached data as it only runs the method if the catalog hasn't yet been built or was built unsuccessfully. Is this sufficient or would it still make sense to implement the lazy-loading strategy as you suggested?

prodonjs: My suggestion was more about not making the `getPriceList` method the public interface from other components since there's really never a case where the entire pricing is shown, but rather the lookup is always based on a particular configuration. In doing this, you could hide the logic that actually retrieves the prices from any other component and make it easier to swap out in the future in the case where we might actually have a dedicated GCP service for returning an exact price based on a customer's project and resource configurations.

jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -202,37 +219,65 @@ export class HardwareScalingForm extends React.Component<Props, State> {
...
+
+    const zone = extractLast(details.instance.zone);
+    const newConfigurationPrice = await priceService.getPrice(
+      zone,
+      configuration
+    );
+    this.setState({
+      newConfigurationPrice,
+    });
+  }
```
yaminahq, line 243: @prodonjs The reason for the significant refactor in this file is due to the fact that the changes in this commit makes getPrice an asynchronous method. In order to separate fetching the data (by awaiting the asynchronous getPrice method) and displaying it I modified the state of this component to include a newConfigurationPrice variable. This variable's value is updated every time the user modifies a value on the form. I also added a class variable oldConfigurationPrice which stores the price estimation of the original configuration and is only updated once when the component is initialized. The price estimation is only displayed if the values of both these variables are not undefined. Any thoughts on my approach?

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,146 @@
...
+/** The monthly price catalog for region and sku */
+type Catalog = Map<string, Map<string, number>>;
+
+export class PriceService {
+  private catalog: Catalog;
+  private _receivedError: boolean;
+
+  private async buildCatalog(): Promise<void> {
+    // Only attempt to build the catalog once
+    if (!this.catalog) {
```
prodonjs, line 34: Just a nit, I'd suggest inverting this and short-circuit right away if this.catalog is populated rather than having to have the entire method body inside of an if block.

jupyterlab_gcedetails/src/service/price_service.ts
```diff
@@ -0,0 +1,146 @@
...
+        this._receivedError = false;
+      } catch (err) {
+        console.log(err);
+        console.error(`Unable to retrieve pricelist.`);
+        this._receivedError = true;
+      }
+    }
+  }
+
+  get receivedError() {
```
prodonjs, line 75: I don't believe this is being used anymore right? If not, I'd just drop it.

prodonjs (dismissed):
Just two small nits but other than that I'm really pleased with how quickly this came to fruition.

PR #298, opened 2020-08-28
[BigQuery] Fix timezone issue with fromtimestamp

CynthiaXJia:
Adds a UTC timezone to `fromtimestamp`, which I didn't realize actually converts to local timezone by default rather than UTC. This will correct the training runs dropdown select dates.


cbwilkes:
Can you fix the merge conflicts?

CynthiaXJia:
> Can you fix the merge conflicts?

Should be good to go

PR #301, opened 2020-08-31
[HardwareConfiguration] Moving service initialization to root

yaminahq:
Currently service initialization happens after the first call to get details from the server as the project name, location, and instance name is needed to set up the services. Moving this to the root index file following the gcp_scheduler widget initialization pattern. If the getMetadata call fails then the error status is passed as props and the widget is launched in an error state.

### DIFF
--- a/jupyterlab_gcedetails/src/index.ts
+++ b/jupyterlab_gcedetails/src/index.ts
@@ -42,8 +42,7 @@ async function activateDetailsWidget(
 
   const clientTransportService = new ClientTransportService();
   const serverProxyTransportService = new ServerProxyTransportService();
-  const detailsUrl = `gcp/v1/details`;
-  const detailsServer = new ServerWrapper(detailsUrl);
+  const detailsServer = new ServerWrapper();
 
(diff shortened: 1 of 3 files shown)


jupyterlab_gcedetails/src/index.ts
```diff
@@ -23,19 +23,72 @@ import {
...
+
+async function activateDetailsWidget(
+  app: JupyterFrontEnd,
+  statusbar: IStatusBar
+) {
+  console.debug('Activating GCP Details Extension');
+
+  const clientTransportService = new ClientTransportService();
+  const serverProxyTransportService = new ServerProxyTransportService();
+  const detailsUrl = `gcp/v1/details`;
```
prodonjs, line 45: I would just inline this since it's not used anywhere else or potentially even hide it inside of the ServerWrapper class itself.

yaminahq: I've made the changes suggested, should be good to merge, just wanted your feedback first that's why I had it as WIP :)

PR #308, opened 2020-09-01
[HardwareConfiguration] Adding tests for HardwareScalingForm

yaminahq:
Adding test file to cover main user flows for the HardwareScalingForm component.

### DIFF
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.spec.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.spec.tsx
@@ -418,12 +418,7 @@ describe('HardwareScalingForm', () => {
     });
   });
 
-  it('Displays correct pricing information', async () => {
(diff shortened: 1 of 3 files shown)


jupyterlab_gcedetails/src/components/hardware_scaling_form.spec.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+    expect(mockOnSubmit).toHaveBeenCalledWith({
+      ...INITIAL_STATE.configuration,
+      attachGpu: true,
+      gpuType: NO_ACCELERATOR_TYPE,
+      gpuCount: NO_ACCELERATOR_COUNT,
+    });
+  });
+
+  it('Displays correct pricing information', async () => {
```
prodonjs, line 421: I believe you could use snapshot tests here rather than the string matching if you'd like.

PR #309, opened 2020-09-01
[HardwareConfiguration] Restricting attaching GPU for incompatible frameworks

yaminahq:
Restricting user from attaching GPUs for instances with an R or Python based framework in the hardware scaling form.

### DIFF
--- a/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
+++ b/jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
@@ -121,8 +121,8 @@ export class HardwareScalingForm extends React.Component<Props, State> {
    */
   private canAttachGpu(machineTypeName: string): boolean {
     const { framework } = this.props.details.instance.attributes;
-    const isValidFramework = GPU_INCOMPATIBLE_FRAMEWORKS.every(
-      incompatibleFramework => !framework.startsWith(incompatibleFramework)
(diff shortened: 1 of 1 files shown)


jupyterlab_gcedetails/src/components/hardware_scaling_form.tsx
```diff
@@ -119,7 +120,12 @@ export class HardwareScalingForm extends React.Component<Props, State> {
    * Currently only N1 general-purpose machines support GPUs: https://cloud.google.com/compute/docs/gpus#restrictions
    */
   private canAttachGpu(machineTypeName: string): boolean {
-    return machineTypeName.startsWith(N1_MACHINE_PREFIX);
+    const { framework } = this.props.details.instance.attributes;
+    const isValidFramework = GPU_INCOMPATIBLE_FRAMEWORKS.every(
```
prodonjs, line 124: I'm pretty sure the framework will be an exact match so you can make `GPU_INCOMPATIBLE_FRAMEWORKS` a Set if you'd like. Or if you want to stick with the array, I might suggest using the `.some()` method and inverting the logic so that you'd have a boolean that's true if any of the incompatible frameworks matched.

yaminahq: When testing I found the R framework for example was fetched as "R:3.6" and thought the versions may differ hence the startsWith logic? I'll invert the logic.

PR #312, opened 2020-09-02
[BigQuery] Dark mode

CynthiaXJia:
dark

Initial version of dark mode. Most colors are set to jp variables (eg `jp-ui-font-color1`) which change automatically when the theme changes. For colors that don't match jp variables (google blue, red, green, and yellow), there is now a `gColor` function that reads the body attribute `data-jp-theme-light` and returns colors according to the [material style color guides](https://standards.google/guidelines/google-material/styles/color/#palettes).

An issue that still needs to be resolved: because we currently aren't listening to theme change signals, the editor will correctly initialize in a particular theme, but if you change the theme once it's loaded, it will not change accordingly. The user would have to open a new editor or refresh the window.

### DIFF
--- a/jupyterlab_bigquery/src/components/details_panel/details_panel.tsx
+++ b/jupyterlab_bigquery/src/components/details_panel/details_panel.tsx
@@ -15,7 +15,7 @@ export const localStyles = stylesheet({
     marginBottom: '10px',
   },
   panel: {
-    backgroundColor: 'white',
+    backgroundColor: 'var(--jp-layout-color0)',
     height: '100%',
     display: 'flex',
     flexDirection: 'column',
@@ -45,12 +45,24 @@ export const localStyles = stylesheet({
   },
 });
 
+interface ChipProps {
+  darkMode: boolean;
+  size?: 'medium' | 'small';
+  component?: any;
+  key?: string | number;
+  label: string;
+}
+
 const StyledChip = withStyles({
   root: {
-    color: '#1967D2',
-    backgroundColor: 'rgba(25, 103, 210, 0.1)',
+    color: (props: ChipProps) =>
+      // white :  blue600
+      props.darkMode ? 'var(--jp-ui-font-color1)' : '#1A73E8',
+    backgroundColor: (props: ChipProps) =>
+      // blue300 at 30% opacity : blue600 at 10% opacity
+      props.darkMode ? 'rgba(138, 180, 248, 0.3)' : 'rgba(26, 115, 232, 0.1)',
   },
-})(Chip);
+})((props: ChipProps) => <Chip {...props} />);
 
 interface SharedDetails {
   id: string;
@@ -103,7 +115,15 @@ export const DetailsPanel: React.SFC<Props> = props => {
                 <div className={localStyles.labelContainer}>
                   {details.labels.map((value, index) => {
                     return (
-                      <StyledChip size="small" key={index} label={value} />
+                      <StyledChip
+                        size="small"
+                        key={index}
+                        label={value}
+                        darkMode={
+                          document.body.getAttribute('data-jp-theme-light') ===
+                          'false'
+                        }
+                      />
                     );
                   })}
                 </div>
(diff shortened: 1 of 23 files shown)


jupyterlab_bigquery/src/components/details_panel/details_panel.tsx
```diff
@@ -45,12 +45,24 @@ export const localStyles = stylesheet({
...
-    backgroundColor: 'rgba(25, 103, 210, 0.1)',
+    color: (props: ChipProps) =>
+      // white :  blue600
+      props.darkMode ? 'var(--jp-ui-font-color1)' : '#1A73E8',
+    backgroundColor: (props: ChipProps) =>
+      // blue300 at 30% opacity : blue600 at 10% opacity
+      props.darkMode ? 'rgba(138, 180, 248, 0.3)' : 'rgba(26, 115, 232, 0.1)',
   },
-})(Chip);
+})((props: ChipProps) => <Chip {...props} />);
```
CynthiaXJia, line 65: For any objects that need to be styled with both the `withStyles` HOC and by reading the body attribute `data-jp-theme-light`, the body attribute or color has to be passed in to the style sheet through props like this.

jupyterlab_bigquery/src/components/shared/bq_table.tsx
```diff
@@ -45,6 +37,55 @@ const localStyles = stylesheet({
...
+    '&$selected': {
+      backgroundColor: 'var(--jp-layout-color2)',
+      '&:hover': {
+        backgroundColor: 'var(--jp-layout-color2)',
+      },
+    },
+    '&:hover': {
+      backgroundColor: 'var(--jp-layout-color2)',
+    },
+  },
```
CynthiaXJia, line 86: Here's where I'm having issues styling the selected `menuItem` within the `TablePagination` component.

cbwilkes: I took this from https://codesandbox.io/s/9t45t?file=/demo.js:478-492. Try changing this to:

```
  menuItem: {
    color: 'var(--jp-ui-font-color1)',
    backgroundColor: 'var(--jp-layout-color0)',
    '&.Mui-selected': {
      backgroundColor: 'var(--jp-layout-color2)',
      '&:hover': {
        backgroundColor: 'var(--jp-layout-color2)',
      },
    },
    '&:hover': {
      backgroundColor: 'var(--jp-layout-color2)',
    },
  },
```

image

cbwilkes (changes requested):

jupyterlab_bigquery/src/components/shared/bq_table.tsx
```diff
@@ -45,6 +37,55 @@ const localStyles = stylesheet({
   },
 });
 
+// TODO: style for dark mode. Currently the disabled buttons are not visible.
+const StyledIconButton = withStyles({
+  root: {
+    color: 'var(--jp-ui-font-color1)',
+    '&$disabled': {
```
cbwilkes, line 44: Also change this to:

```
const StyledIconButton = withStyles({
  root: {
    color: 'var(--jp-ui-font-color1)',
    '&.Mui-disabled': {
      color: 'var(--jp-ui-font-color3)',
    },
  },
})(IconButton);
```

cbwilkes (changes requested):

cbwilkes:
Also, please fix the merge conflicts when you get a chance

jupyterlab_bigquery/src/components/shared/bq_table.tsx
CynthiaXJia: Awesome! Thanks!

cbwilkes (changes requested):
Having `document.body.getAttribute('data-jp-theme-light')` at various locations seems difficult to track and could change in the future. Can you make a central utility function like `IsDarkTheme()` which does the `document.body.getAttribute('data-jp-theme-light') === 'false'` check?

This way if the dark mode detection logic changes in the future, we can control it from a single location.

cbwilkes (approved):

PR #314, opened 2020-09-02
[HardwareConfiguration] Fixing missing framework bug

yaminahq:
Before this PR the extension would crash on the hardware scaling form when doing the framework-GPU compatibility check if a framework wasn't fetched with the GCE metadata. In production this shouldn't be an issue as we assume a framework should always be specified but for the sake of local testing I've added a fix that assumes the framework is compatible with attaching GPUs if a framework isn't specified.


prodonjs (approved):
LGTM. I don't see any other things I'd obviously want to account that we discussed in our standup today. This seems sufficient for now and if we find other things, we can change them.

PR #317, opened 2020-09-03
[BigQuery] Snackbar with autohide when copying

CynthiaXJia:
Adds a snackbar popup when a user copies from the context menu or from the query editor top bar. 

Alters the snackbar component - now you can pass in a duration in ms before the snackbar automatically disappears. The one that pops up after you copy something lasts for 2 seconds.

### DIFF
--- a/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
+++ b/jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
@@ -37,6 +37,7 @@ import { formatBytes } from '../../../utils/formatters';
 import QueryResultsManager from '../../../utils/QueryResultsManager';
 import { isDarkTheme } from '../../../utils/dark_theme';
 import { SnackbarState, openSnackbar } from '../../../reducers/snackbarSlice';
+import { COPIED_AUTOHIDE_DURATION } from '../../shared/snackbar';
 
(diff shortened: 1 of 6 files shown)


jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
```diff
@@ -628,6 +631,10 @@ class QueryTextEditor extends React.Component<
         onClick={_ => {
           const query = this.editor.getValue();
           copy(query.trim());
+          this.props.openSnackbar({
+            message: 'Query copied',
+            autoHideDuration: 2000,
```
cbwilkes, line 636: Can you make this a const in a central location? Like `const AUTO_HIDE_DURATION = 2000`?

cbwilkes (changes requested):

jupyterlab_bigquery/src/components/query_editor/query_text_editor/query_text_editor.tsx
CynthiaXJia: Yes! I added a `const COPIED_AUTOHIDE_DURATION = 2000` in `snackbar.tsx` and moved `snackbar.tsx` to `src/shared/snackbar`

cbwilkes (approved):

PR #322, opened 2020-09-08
[BigQuery] Add python docstrings to details and query history handlers

CynthiaXJia:


cbwilkes (approved):
Thanks for adding comments

PR #336, opened 2020-09-11
[BigQuery] remove passing dataframe logic via front-end

MRzNone:


jupyterlab_bigquery/jupyterlab_bigquery/query_incell_editor.py
```diff
@@ -22,16 +21,9 @@ class QueryIncellEditor(DOMWidget):
 
   query = Unicode().tag(sync=True)
   query_flags = Unicode().tag(sync=True)
-  result = Any().tag(sync=True)
 
 
 @magic_arguments.magic_arguments()
-@magic_arguments.argument(
```
cbwilkes, line 29: Does the `%bigquery_editor` magic fail when this argument exists?

We prefer to maintain flag capability with `%bigquery` even if it mean ignoring `%bigquery` flags.

MRzNone: I added the unsupported args and it will inform users if there there is unsupported args.  The editor will display normally when there are unsupported falgs.

cbwilkes (approved):

PR #337, opened 2020-09-11
[BigQuery] change version number

MRzNone:


jupyterlab_bigquery/setup.py
```diff
@@ -55,5 +55,8 @@
         "google-cloud-bigquery~=1.25.0",
```
cbwilkes, line 55: Can you convert all of these from `~=` to `>=`? we had an issue with dependencies downgrading.

PR #349, opened 2020-09-16
[BigQuery] UX updates

MRzNone:
- change bigquery icon
- append dataframe code when in notebook, and added snackbar pop-up

image


MRzNone:
Synced off-line with UX designer about dataframe code generation.
- If we are in tab editor, we copy the code
- if we are in notebook, we append the code

cbwilkes (approved):

PR #355, opened 2020-09-19
Update README.md for user instructions

gogasca:


gogasca:
Only available to google.com

PR #356, opened 2020-09-23
Changed shared components and other buttons to use material styles

mounikasomisetty:
Changed selectInput, textInput, actionBar, submitButton etc to use material styles. Added a badge component based on material's Chip component.
Also changed tests to reflect the dom structure changes of the above components.

[Badge component screenshot](http://screen/4Vk2fgkgehAWGzS)
[Scheduler Form w/ new shared components screenshot](http://screen/AKHHTvn36Tifq4q)


shared/src/components/badge.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { COLORS, BASE_FONT } from '../styles';
+
+interface BadgeProps {
+  value?: string;
+}
+
+const StyledChip = withStyles({
+  root: {
+    color: COLORS.white,
+    backgroundColor: '#555555',
```
prodonjs, line 29: We've been trying to support both dark and light theme modes. As such, we've tried to rely on the JupyterLab color variables as much as possible rather than fixed colors.

COLORS.white already handles this. I think we might be able to use one of the jp-inverse-layout-color variables for this perhaps.

https://github.com/jupyterlab/jupyterlab/blob/1.2.x/packages/theme-light-extension/style/variables.css

shared/src/styles.ts
```diff
@@ -47,6 +47,12 @@ export const BASE_FONT: types.CSSProperties = {
   fontSize: 'var(--jp-ui-font-size1, 13px)',
 };
 
+export const INPUT_TEXT_STYLE = {
```
prodonjs, line 50: You could also use the spread syntax here.

```
export const INPUT_TEXT_STYLE = {
  ...BASE_FONT,
  color: 'var(--jp-ui-font-color1)',
};
```

prodonjs (commented):
Overall this looks great. See my inline suggestions as well as these points.

1. Please make sure to increment the version numbers in the shared/ and jupyterlab_gcpscheduler/ folders.
  - You will need to do this in the package.json and Python package files.
2. I think it would also be helpful to add the Alpha badging to the Scheduler dialog in addition to the sidebar component. Can we try adding that next to the dialog's title?

Other than that, this looks fantastic.

shared/src/styles.ts
mounikasomisetty: It really doesn't like the type conversion there, so it seems a lot simpler to keep it like this

shared/src/components/badge.tsx
mounikasomisetty: Ah gotcha. Testing out that change!

shared/src/styles.ts
prodonjs: SGTM

mounikasomisetty:
> Overall this looks great. See my inline suggestions as well as these points.
> 
> 1. Please make sure to increment the version numbers in the shared/ and jupyterlab_gcpscheduler/ folders.
> 
> * You will need to do this in the package.json and Python package files.
> 
> 1. I think it would also be helpful to add the Alpha badging to the Scheduler dialog in addition to the sidebar component. Can we try adding that next to the dialog's title?
> 
> Other than that, this looks fantastic.

Screenshot for dialog title https://screenshot.googleplex.com/4GX6NbXnz7qXU8B

prodonjs:
@mounikasomisetty - Can you update the second screenshot with how the dialog looks? Also, if you can take one with dark mode on that would be terrific as well? I'm going to share with Huaiwei to get his approval.

prodonjs (approved):

PR #357, opened 2020-09-28
[HWConfig] Simplification refactoring and adds Alpha badge

prodonjs:
- Exports individual classes from hardware config for alternative extension structure.

Screenshot 1
Screenshot 2


prodonjs:
@cbwilkes This is a really large change but mostly it's just refactoring to allow this to work in the managed environment. PTAL when you get a chance but don't feel obligated to do a fine-grained review.

cbwilkes (approved):

PR #364, opened 2020-10-08
Rename Optimizer extension to Vizier as per documentation

angelakuo:
https://cloud.google.com/ai-platform/optimizer/docs

Not sure if/when they will change the URL but the API itself has been renamed to Vizier


prodonjs (approved):
Looks good!

PR #369, opened 2020-10-15
Use the same region us-central1 for all Vizier API calls regardless o…

angelakuo:
…f instance region. This is because Vizier is only available in that particular region

### DIFF
--- a/jupyterlab_vizier/src/service/vizier_service.ts
+++ b/jupyterlab_vizier/src/service/vizier_service.ts
@@ -54,17 +54,11 @@ export function prettifyTrial(rawTrialName: string): string {
  * Class to interact with Vizier
  */
 export class VizierService {
+  // Only available in this region regardless of instance zone
+  private readonly REGION = 'us-central1';
   private metadataPromise?: Promise<InstanceMetadata>;
 
   constructor(private _transportService: TransportService) {}
-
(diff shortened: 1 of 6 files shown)


jupyterlab_vizier/src/utils/urls.ts
```diff
@@ -1,163 +1,140 @@
 import {
   fakeProjectId,
-  fakeRegion,
   cleanFakeStudyName,
   cleanFakeTrialName,
   fakeCleanOperationName,
 } from '../service/test_constants';
 
+const REGION = 'us-central1';
```
prodonjs, line 8: Perhaps export this so it could be shared with the VizierService.

jupyterlab_vizier/src/service/vizier_service.ts
```diff
@@ -54,18 +54,12 @@ export function prettifyTrial(rawTrialName: string): string {
  * Class to interact with Vizier
  */
 export class VizierService {
+  // Only available in this region regardless of instance zone
+  private readonly REGION = 'us-central1';
```
prodonjs, line 58: Suggest assigning this value from the shared constant or using it directly if you don't expect it to ever be an instance level value.

prodonjs (changes requested):

prodonjs (approved):
LGTM.

PR #377, opened 2020-11-17
Initial commit for realtime collaborative editing

itsazhuhere:
This commit is for the Jupyter realtime collaboration extension that I am working on as an engineering resident.
This commit is also the first part of a larger commit, broken up for clarity, so if any comments or such refer to implementation that isn't currently there, later commits will bring more context. The focus of this commit is to establish a base extension using the cookiecutter and to introduce a class that handles applying remote changes to the local notebook.


prodonjs:
Closing this per conversation.

PR #382, opened 2020-11-24
Scheduler UX changes

mounikasomisetty:
Screencast: http://go/scrcast/NTgxODEwOTk1OTA3Nzg4OHxkZTE5NTQ1Ny1kZA

*Testing for pagination, cloud bucket component and share dialog remaining


prodonjs (commented):
I love the UX. Comments mostly centered around the following:

- Defining anonymous functions inside `render`. 
  - If possible prefer private methods [bound in constructor](https://reactjs.org/docs/handling-events.html) for succinctness
- Simplified interfaces for GCS types
  - If we don't directly access a field, you don't need to include it in the interface definition since none of that is kept at runtime
- Sidebar icon
  - You probably want to include these [icon changes](https://github.com/GoogleCloudPlatform/jupyter-extensions/commit/c0c46d89faab90d84b10caffe8104e60dad30483#diff-e57041e991963b5236608a88cd36938632f2d42fe480340a88644aba83098acf)

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
```diff
@@ -137,20 +219,73 @@ export class GcpScheduledJobsPanel extends React.Component<Props, State> {
...
   }
 
-  private async _getJobs() {
+  private async _getRunsAndSchedules(pageSize: number, pageToken?: string) {
     try {
       this.setState({ isLoading: true, error: undefined });
-      const jobs = await this.props.gcpService.listNotebookJobs();
-      this.setState({ isLoading: false, jobs });
+      if (!pageToken) {
+        const runs = await this.props.gcpService.listRuns(pageSize);
```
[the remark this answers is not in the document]
mounikasomisetty: Yeah I was considering that, I can make that change.

jupyterlab_gcpscheduler/src/components/cloud_bucket.tsx
```diff
@@ -0,0 +1,243 @@
+/* eslint-disable no-use-before-define */
```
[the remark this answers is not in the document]
mounikasomisetty: I have a bug to remove all these object definitions, so I can follow up with another pull request for this.

jupyterlab_gcpscheduler/src/components/job_list_item.tsx
```diff
@@ -13,163 +13,186 @@
...
+    super(props);
+    this.state = { anchorEl: null };
+  }
+
+  render() {
+    const { gcpService, job } = this.props;
+    const schedule = 'schedule' in job;
+    const endTime = new Date(job.endTime || job.createTime);
+
+    const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
```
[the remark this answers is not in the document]
mounikasomisetty: Ah gotcha.

jupyterlab_gcpscheduler/src/components/cloud_bucket.tsx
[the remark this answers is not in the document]
mounikasomisetty: Ah didn't notice that, copy paste issues!

jupyterlab_gcpscheduler/src/service/gcp.ts
```diff
@@ -356,4 +432,63 @@ export class GcpService {
...
+      ...this.createJob(job, projectId),
+      type,
+      bucketLink,
+    };
+  }
+
+  private createSchedule(job: AiPlatformJob, projectId: string): Schedule {
+    return {
+      ...this.createJob(job, projectId),
+      schedule: '30 12 */2 * *',
```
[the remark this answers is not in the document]
mounikasomisetty: Added

shared/src/styles.ts
```diff
@@ -39,7 +39,8 @@ export const COLORS = {
   link: 'var(--jp-content-link-color, #3367d6)',
   red: 'var(--jp-error-color1, #d93025)',
   white: 'var(--jp-layout-color1, white)',
-  inverse: 'var(--jp-ui-inverse-font-color2, #555555)',
+  inverse: 'var( --jp-ui-font-color3, #555555)',
```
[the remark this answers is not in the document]
mounikasomisetty: Fixed

jupyterlab_gcpscheduler/src/interfaces.ts
```diff
@@ -0,0 +1,401 @@
...
+}
+export interface CloudStorageApiBuckets {
+  /** The list of items. */
+  items?: CloudStorageApiBucket[];
+  /** The kind of item this is. For lists of buckets, this is always storage#buckets. */
+  kind?: string;
+  /** The continuation token, used to page through large result sets. Provide this value in a subsequent request to return the next page of results. */
+  nextPageToken?: string;
+}
+export interface Job {
```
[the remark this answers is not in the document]
mounikasomisetty: The Job/Run/Schedule interfaces are intermediary interfaces similar to JobRow in pantheon, so that when we do replace the API we can just change the createJob, createRun and createSchedule methods.

jupyterlab_gcpscheduler/src/interfaces.ts
```diff
@@ -0,0 +1,401 @@
...
+export interface Bucket {
+  name: string;
+  accessLevel?: 'uniform' | 'fine';
+}
+
+export interface Buckets {
+  buckets: Bucket[];
+}
+
+export interface CloudStorageApiBucket {
```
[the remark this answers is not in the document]
mounikasomisetty: Done

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -88,6 +84,7 @@ interface Props {
 interface State {
   dialogClosedByUser: boolean;
   gcpSettings?: GcpSettings;
+  metadataProjectId?: string;
```
prodonjs, line 87: Correct me if I am wrong, but I believe you introducing this because we will no longer have the initialization component that was responsible for setting the projectId in settings? 

Do we need to have this additional state variable if we would just move the logic that retrieves projectId and sets it to this module? Where would projectId ultimately be set in the GcpSettings?

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
```diff
@@ -137,59 +154,55 @@ export class GcpScheduledJobsPanel extends React.Component<Props, State> {
   componentDidUpdate(prevProps: Props) {
     if (
       this.props.isVisible &&
-      !(prevProps.isVisible || this.state.isLoading)
+      !(prevProps.isVisible || this.state.runsTab.isLoading)
     ) {
-      this._getRunsAndSchedules(this.state.rowsPerPage);
+      this._getRuns(this.state.rowsPerPage);
+      this.setState({ tab: 0, page: 0 });
```
prodonjs, line 160: Could we leave the user on the tab they were previously on but set the page to zero? That way, if they created a new Schedule, they would see it appear if they moved from the file browser to the Executor panel.

shared/gcp_jupyterlab_shared/version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION = '1.0.11'
+VERSION = '1.0.13'
```
prodonjs, line 15: Any reason we are bumping it by 2 minor versions?

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -164,10 +167,9 @@ export class SchedulerDialog extends React.Component<Props, State> {
   }
 
   render() {
-    const { gcpSettings } = this.state;
-    const projectId = gcpSettings ? gcpSettings.projectId : '';
+    const projectId = this.getProjectId();
     return (
-      <Dialog open={this._isOpen()}>
+      <Dialog open={this._isOpen()} scroll="body">
```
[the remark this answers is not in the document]
mounikasomisetty: The difference is best illustrated here https://material-ui.com/components/dialogs/#scrolling-long-content

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: So for me, gcpSettings was cleared so I noticed this was necessary. I thought it was possible other users may still have it and it could be a good fallback

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
mounikasomisetty: Ah okay. Can fix that

shared/gcp_jupyterlab_shared/version.py
mounikasomisetty: I think I forgot to at some point change this file, package.json was at 1.0.13

jupyterlab_gcpscheduler/src/components/cloud_bucket.tsx
```diff
@@ -240,4 +166,86 @@ export class CloudBucketSelector extends React.Component<Props, State> {
       />
     );
   }
+
+  private async _getBuckets(selectedBucketName?: string) {
+    try {
+      this.setState({ isLoading: true, error: undefined });
+      const buckets = await this.props.gcpService.listBuckets();
+      const validBucketOptions = buckets.buckets.filter(
```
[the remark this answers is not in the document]
mounikasomisetty: Done

jupyterlab_gcpscheduler/src/components/cloud_bucket.tsx
```diff
@@ -63,139 +61,66 @@ export class CloudBucketSelector extends React.Component<Props, State> {
...
-        this.createAndSelectBucket(newBucketName);
-      }
-      this.props.onGcsBucketChange(newBucketName);
+  handleGetOptionLabel(option: string | BucketOption) {
+    // Value selected with enter, right from the input
+    if (typeof option === 'string') {
+      return option;
     }
+    // Add "xxx" option created dynamically
+    if ((option as BucketOption).inputValue) {
```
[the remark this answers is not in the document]
mounikasomisetty: Done

jupyterlab_gcpscheduler/src/components/dialog.tsx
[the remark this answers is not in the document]
prodonjs: I see. I actually thought it might be the "paper" behavior where only the body of the dialog scrolls but the overall document that it appears in doesn't. How does this actually look when you're submitting a job?

shared/gcp_jupyterlab_shared/version.py
prodonjs: Ok. I was just wondering what happened to 1.0.12 but then I realized that it was a previous change in this branch that is not yet on master.

jupyterlab_gcpscheduler/src/components/dialog.tsx
prodonjs: I guess my question is around where you would expect the `projectId` setting to be saved? Since we used to do it in the Initialization component that is no longer present, I think it would be best if somehow the logic inside of the `_settingsChanged` method would be to check to see if `projectId` was set, and if not, retrieve it from the `gcpService` property.

In the current implementation, it looks as if you'd be retrieving it from the server each time but never actually storing it in settings. That appears to be why you have to pass `projectId` into SchedulerForm explicitly rather than rely on the setting property to be present.

jupyterlab_gcpscheduler/src/components/dialog.tsx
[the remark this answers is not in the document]
mounikasomisetty: http://go/scrcast/NTc3ODgwNDkwMTQxMjg2NHwzNjBhN2RmOC0zZQ

I think it looks a lot cleaner than the paper, which make the form seem cramped.

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: Ah I see what you mean. initially I thought we could just phase out the gcpSettings.projectId, but I can use it.

mounikasomisetty: Where do you think the gcpSettings.projectId should be set then?

jupyterlab_gcpscheduler/src/components/dialog.tsx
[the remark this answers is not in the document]
prodonjs: I agree. Let's roll with it for now and if UX has thoughts otherwise, we can pivot later. I'm starting to wonder if our form is going to be too big to justify a dialog and instead require a new tab.

jupyterlab_gcpscheduler/src/components/dialog.tsx
prodonjs: It's your call. The original intent of `GcpSettings.projectId` was to facilitate the idea of using the extension _outside_ a GCE VM, but that's no longer a goal.

The only disadvantage I see is that we will always be introducing an extra call to the backend when the extension is added to the DOM.

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -263,11 +250,20 @@ export class SchedulerDialog extends React.Component<Props, State> {
...
       this.props.projectStateService.projectId = settings.projectId;
+    } else {
+      let projectId = undefined;
+      try {
+        projectId = await this.props.gcpService.projectId;
+      } catch (err) {
+        console.info(`${err} occurred when trying to get projectId`);
+      }
+      this.props.projectStateService.projectId = projectId;
+      settings.projectId = projectId;
```
prodonjs, line 266: Per our Chat conservation, if we want to continue to persist the projectId in the settings, we should add a call to `this.props.settings.set('projectId'...` similar to https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/e534ba1bbb80b332419210af4421d3979db94653/jupyterlab_gcpscheduler/src/components/dialog.tsx#L298.

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -263,11 +250,20 @@ export class SchedulerDialog extends React.Component<Props, State> {
...
     const settings = (newSettings.composite as unknown) as GcpSettings;
     if (settings.projectId) {
       console.info(`Using ${settings.projectId} for GCP API calls`);
       this.props.projectStateService.projectId = settings.projectId;
+    } else {
+      let projectId = undefined;
+      try {
+        projectId = await this.props.gcpService.projectId;
+      } catch (err) {
+        console.info(`${err} occurred when trying to get projectId`);
```
prodonjs, line 263: I think this is pretty much a fatal condition, albeit one that shouldn't occur in practice. I'm wondering if we should just let it error out which would cause the extension not to load.

prodonjs (changes requested):

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: Done

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: can do

prodonjs (approved):

PR #386, opened 2020-12-04
[BigQuery] Copy to DataFrame should support multiline queries

cbwilkes:
image


prodonjs (approved):

prodonjs:
Hmm, not sure why the build failed.

prodonjs:
Hmm it seems to be recurring consistently which is peculiar since `postpush` ran successfully yesterday. Unfortunately I don't think it's possible to get the error log from the job failure that is mentioned.

One thing I would try is to try to execute the build process locally.

prodonjs:
I opened #387 to address the presubmit issue independently of this change. If you find a solution in the meantime that works for this, please include it here and I'll drop the other one.

https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/387#issuecomment-739898516

prodonjs:
OK, if you rebase this change over master, presubmits should pass and we can get this into the build.

prodonjs (approved):

PR #387, opened 2020-12-07
Fix scripts to resolve presubmit issue

prodonjs:
- The root causes appears to be an issue with Webpack transpiling the `highlight.js` module which is an indirect dependency of the uCAIP extension.
- Because this is not currently being offered for automated installation, we'll remove this temporarily from the `build_jupyterlab.sh` script.


prodonjs:
Able to replicate the issue locally. Attaching build file.

[build_failure.txt](https://github.com/GoogleCloudPlatform/jupyter-extensions/files/5652724/build_failure.txt)

cbwilkes (approved):

PR #389, opened 2020-12-08
Revert div to fragment for BQTable scrolling; Update to bq 0.1.5

cbwilkes:
image


cbwilkes:
Fragment React document https://reactjs.org/docs/fragments.html

prodonjs (approved):
Ouch I'm the one that broke this. I didn't realize that but thanks for finding it.

PR #390, opened 2020-12-08
Scheduler environment labelling changes

mounikasomisetty:
1. Unified the labeling used between Scheduler and Pantheon (ie. Environment)
2. Ensured that all container images are up-to-date and consistent with the Pantheon options (except Swift)
3. Replaced the key that the Runtime handler looks at to use the ENV_URI_FILE_PATH, and accordingly changed logic in the getImageUri functions
4. Added custom container functionalities
5. Change scroll on dialog to match mock http://slides/1Ce2_tF7xIV9_A-40dqbT0coejEAGRjmcwqmwysz6eSQ#slide=id.ga5acc72875_0_5
6. Change error style to match mock http://slides/1Ce2_tF7xIV9_A-40dqbT0coejEAGRjmcwqmwysz6eSQ#slide=id.ga5acc72875_0_0

Screencast: http://go/scrcast/NTgyNTcxMDg0MzIzMjI1Nnw5OTMxOTRiZC03NA


jupyterlab_gcpscheduler/src/data.ts
```diff
@@ -99,19 +103,46 @@ export const CONTAINER_IMAGES: Option[] = [
...
     value: `${GCR_PREFIX}/r-cpu.3-6:latest`,
     text: 'R 3.6 (with r-essentials)',
   },
+  {
+    value: `${GCR_PREFIX}/base-cpu:latest`,
+    text: 'Python 2 and 3',
+  },
   { value: `${GCR_PREFIX}/beam-notebooks:latest`, text: 'Apache Beam' },
+  {
+    value: 'gcr.io/cloud-dataproc/dataproc-spawner:prod',
```
prodonjs, line 136: I don't think we should actually be supporting the Dataproc Spawner image. It's not actually a Notebook image but an image used to spawn a Dataproc Cluster which then hosts JupyterLab.

https://cloud.google.com/dataproc/docs/tutorials/dataproc-hub-users

shared/gcp_jupyterlab_shared/handlers.py
```diff
@@ -30,7 +30,7 @@
     'http://metadata.google.internal') + '/computeMetadata/v1/?recursive=true'
 METADATA_HEADER = {'Metadata-Flavor': 'Google'}
 SCOPE = ('https://www.googleapis.com/auth/cloud-platform',)
-FRAMEWORK_ENV_VAR = 'ENV_VERSION_FILE_PATH'
+ENV_URI_VAR = 'ENV_URI_FILE_PATH'
```
prodonjs, line 33: FYI, you'll also need to introduce this change in the DLVM server extension repository at https://dlvm.googlesource.com/binaries-build/+/refs/heads/master/packages/server_extension/caip_notebooks_serverextension/common/handlers.py#209

shared/src/components/message.tsx
```diff
@@ -31,42 +31,41 @@ interface Props {
 const localStyles = stylesheet({
   error: {
     backgroundColor: 'var(--md-red-50, #ffebee)',
```
prodonjs, line 33: I'm wondering if there's a better background color based on the CSS variables at https://github.com/jupyterlab/jupyterlab/blob/1.2.x/packages/theme-light-extension/style/variables.css?

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -169,36 +174,57 @@ export class SchedulerDialog extends React.Component<Props, State> {
...
-                    target="_blank"
-                    onClick={menuCloseHandler}
-                  >
-                    View all scheduled executions
-                  </a>
-                </SmallMenuItem>,
-                <SmallMenuItem
-                  key="reset"
-                  onClick={() => this._onResetSettings(menuCloseHandler)}
+            <IconButton onClick={this.handleClick}>
```
prodonjs, line 196: Was there an issue with modifying the items passed to the shared `IconButtonMenu` component?

jupyterlab_gcpscheduler/src/data.ts
mounikasomisetty: Removed

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: Ahhh, no, I just wanted the styles to be consistent with the sidebar Menu as well.

shared/src/components/message.tsx
mounikasomisetty: http://b/175132231

shared/gcp_jupyterlab_shared/handlers.py
mounikasomisetty: http://b/175132231

shared/src/components/message.tsx
mounikasomisetty: Changed to  --jp-error-color3: var(--md-red-100);

jupyterlab_gcpscheduler/src/components/dialog.tsx
prodonjs: Ahh I see. It looks like there's a lot of boilerplate added in both places for some of the click handling? Would it be possible to factor the new styles out to the shared `IconButtonMenu` component? It looks like Scheduler is the only extension to use it anyways.

mounikasomisetty: Ooo okay. Can do that!

mounikasomisetty: Done!

prodonjs (approved):

PR #393, opened 2020-12-10
Add ssl option PROTOCOL_TLS to tornado HTTP requests

cbwilkes:
This will allow SSL to succeed on certain systems like Macs while continuing to succeed on currently working systems

### DIFF
--- a/shared/gcp_jupyterlab_shared/handlers.py
+++ b/shared/gcp_jupyterlab_shared/handlers.py
@@ -14,6 +14,7 @@
 """Tornado request handler classes for extension."""
 import json
 import os
+import ssl
 from base64 import b64decode
 
 import google.auth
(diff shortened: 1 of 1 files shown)


shared/gcp_jupyterlab_shared/handlers.py
```diff
@@ -31,11 +32,14 @@
 METADATA_HEADER = {'Metadata-Flavor': 'Google'}
 SCOPE = ('https://www.googleapis.com/auth/cloud-platform',)
 FRAMEWORK_ENV_VAR = 'ENV_VERSION_FILE_PATH'
-
+SSL_OPTIONS = {"ssl_version": ssl.PROTOCOL_TLSv1}
```
prodonjs, line 35: https://docs.python.org/3.6/library/ssl.html#ssl.PROTOCOL_TLSv1 suggests that this is deprecated. We have had this library supporting Scheduler since last year on Notebooks VMs and haven't had any issues.

I'm somewhat reluctant to introduce it to fix a local development issue if it could potentially affect production behavior.

cbwilkes:
I set it to PROTOCOL_TLS which works on both systems and is not deprecated https://docs.python.org/3.6/library/ssl.html#ssl.PROTOCOL_TLS

prodonjs:
Alright let's roll with that. I will rebase this into my outstanding [PR](https://github.com/GoogleCloudPlatform/jupyter-extensions/pull/394) so that the change is included in the next shared release.

prodonjs (approved):

PR #398, opened 2020-12-11
Allow public datasets to be searched via datacatalog api

cbwilkes:
Allow Data Catalog to search public datasets


prodonjs (approved):
There's nothing I love more than enabling a feature via a one-liner!

PR #401, opened 2021-01-07
Proto changes, UX fixes

mounikasomisetty:
Dialog: http://screen/BkSENuzaVKBtjta
Schedule Components: http://screen/8AApJKDup66dd8t, http://screen/jHMiwdZiHfTgwpm, http://screen/7A2E27RNewMtbd6, http://screen/YNAAvex7dXjBqFT
Submitted Job: http://screen/AZXwAV3CnnM7okC
Sidebar: http://screen/59F2BqrV4MTzzxk, http://screen/BnKKL38pUArwUHq


prodonjs (approved):
No concerns here. I suggest that we try to get started on integrating the `new-scheduler` branch into `master`.

1. First, let's get all changes to the `shared/` packages in a PR. These will also need to be upstreamed to the DLVM image repository.
2. Next, we'll want to create the `jupyterlab_executor` package in the root of the repository.

jupyterlab_gcpscheduler/src/data.ts
```diff
@@ -77,68 +82,104 @@ export function removeFromList<T>(list: T[], value: T) {
...
-  { value: `${GCR_PREFIX}/base-cu100:latest`, text: 'CUDA Toolkit 10.0' },
+  {
+    value: 'gcr.io/kaggle-gpu-images/python:latest',
+    text: 'Kaggle (GPU)',
+    searchKeywords: ['gcr.io', 'kaggle', 'gpu'],
+  },
+  {
+    value: 'gcr.io/kaggle-images/python:latest',
+    text: 'Kaggle (CPU)',
+    searchKeywords: ['gcr.io', 'kaggle-images'],
```
prodonjs, line 166: Just double-checking here that the Kaggle CPU image follows a different convention that the GPU one.

jupyterlab_gcpscheduler/src/data.ts
```diff
@@ -77,68 +82,104 @@ export function removeFromList<T>(list: T[], value: T) {
   }
 }
 
-export const CUSTOM_CONTAINER: Option = {
+export const CUSTOM_CONTAINER: Framework = {
   value: 'customContainer',
   text: 'Custom Container',
+  searchKeywords: [] as string[],
```
prodonjs, line 88: Do you need the `[] as string[]` here? I would have though the explicit interface definition would suffice.

jupyterlab_gcpscheduler/src/components/scheduler_form.tsx
```diff
@@ -181,14 +181,16 @@ export class InnerSchedulerForm extends React.Component<
...
           : this.props.values.imageUri;
-        const lastColonIndex = imageUri.lastIndexOf(':');
-        const matchImageUri =
-          lastColonIndex !== -1 ? imageUri.substr(0, lastColonIndex) : imageUri;
-        const matched = ENVIRONMENT_IMAGES.find(i => {
-          return String(i.value).startsWith(matchImageUri);
+        if (imageUri === CUSTOM_CONTAINER.value) {
+          imageUri = this.props.values.customContainerImageUri;
+        }
+        const matched = ENVIRONMENT_IMAGES.slice(1).find(env => {
```
prodonjs, line 190: Correct me if I am mistaken, but it looks like the algorithm here is to loop through each environment image until you find one where all keywords are present in the `imageUri`? If that's the case, it might be slightly more readable to use a traditional `for ... of` loop through `ENVIRONMENT_IMAGES`. Then, you can use something like `env.searchKeywords.every(k => String(imageUri).includes(k))` and if true you can break the loop.

It's probably just me but the negation of the `env.searchKeywords.find` and the include inside mixed me up a bit.

prodonjs:
@mounikasomisetty just circling back on this one. I am OK with you merging as-is but just wanted to bring your attention to a few things from my comments.

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -170,54 +158,21 @@ export class SchedulerDialog extends React.Component<Props, State> {
...
   render() {
-    const projectId = this.getProjectId();
     return (
       <Dialog open={this._isOpen()} scroll="body">
         {this.state.showCreateForm && (
           <header className={localStyles.header}>
             <span className={localStyles.title}>
-              Create a notebook{' '}
-              {this.state.creatingExecution ? 'execution' : 'schedule'}{' '}
+              Submit notebooks to Executor
```
prodonjs, line 166: *nit* Should this be notebook since we're only submitting one file for execution at a time?

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
```diff
@@ -97,9 +97,20 @@ const localStyles = stylesheet({
   },
 });
 
+const StyledTabs = withStyles({
+  indicator: {
+    backgroundColor: 'rgba(0,0,0,0)',
```
prodonjs, line 102: Should we try to use one of the JupyterLab variable colors to protect us against having to address this if/when we want to support a dark theme?

jupyterlab_gcpscheduler/src/components/scheduler_form.tsx
```diff
@@ -127,11 +126,11 @@ function getName(notebookName: string) {
 
 const localStyles = stylesheet({
   scroll: {
-    maxHeight: '52vh',
+    maxHeight: '56vh',
     overflowY: 'scroll',
     overflowX: 'hidden',
-    borderTop: '0px solid #777',
-    borderBottom: '0px solid #777',
+    borderTop: '1px solid #DADCE0',
```
prodonjs, line 132: Same as above, should we be trying to use the CSS color variables from https://github.com/jupyterlab/jupyterlab/blob/master/packages/theme-light-extension/style/variables.css where possible to avoid theming conflicts?

shared/src/components/checkbox_input.tsx
```diff
@@ -37,18 +37,28 @@ const containerStyle = style({
...
+
 /** Funtional Component for Checkbox input fields */
 export function CheckboxInput(props: CheckboxInputProps) {
   const { label, checked, ...inputProps } = props;
   return (
     <div className={containerStyle}>
       <StyledLabel
         control={
-          <Checkbox
+          <CustomColorCheckBox
```
prodonjs, line 61: I assume this _shouldn't_ have any noticeable affect on other extensions that may be using this component?

prodonjs (commented):
Just a few small comments. In the future, if possible please try to open a separate PR for changes unrelated to the original one which centered around the API interface. That will help make reviews and merges simpler. Please update the title of this PR.

shared/src/components/checkbox_input.tsx
mounikasomisetty: Yes it's just a color change. But I checked and made sure no other extensions are using this component.

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
mounikasomisetty: Ah this is me overriding the default indicator (which had a responsiveness issue) and setting it to transparent (the 0 alpha value)

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: I get your point. I'll change it and let Huaiwei know.

jupyterlab_gcpscheduler/src/components/scheduler_form.tsx
mounikasomisetty: Can do. I was trying to use the color from Figma but I think I found a jp alternative.

shared/src/components/checkbox_input.tsx
mounikasomisetty: Correct. Also it's not used in any other extension.

jupyterlab_gcpscheduler/src/data.ts
mounikasomisetty: Yeah, unfortunately.

jupyterlab_gcpscheduler/src/data.ts
mounikasomisetty: Done. It must have been an IDE delay, I had thought it threw a lint error.

jupyterlab_gcpscheduler/src/components/jobs_panel_widget.tsx
prodonjs: Gotcha!

jupyterlab_gcpscheduler/src/components/scheduler_form.tsx
prodonjs: Excellent. Please resolve once the change is made.

jupyterlab_gcpscheduler/src/components/dialog.tsx
```diff
@@ -170,54 +158,21 @@ export class SchedulerDialog extends React.Component<Props, State> {
...
               <Badge value="alpha" />
             </span>
-            <IconButtonMenu
```
prodonjs, line 183: Forgot to comment on this this morning, is there a reason we're moving out all of this? I know in the past the Reset configuration was useful to clear out settings that might have mistakenly been saved and this would also drop the View link?

jupyterlab_gcpscheduler/src/components/scheduler_form.tsx
mounikasomisetty: Fixed.

jupyterlab_gcpscheduler/src/components/dialog.tsx
mounikasomisetty: Huaiwei asked me if we could drop it. I can add it back if you feel it's useful.

prodonjs: I'm fine with dropping the links to the Pantheon pages, but for now let's leave the Reset settings since it's an easy way to clear out anything that might have been previously saved that might be affecting the initial dialog state.

mounikasomisetty: That's fair. Will add it back

prodonjs (approved):
LGTM to me. Feel free to merge after the presubmit finishes.

PR #403, opened 2021-01-12
[BigQuery] Use server proxy to list tables and models

cbwilkes:


jupyterlab_bigquery/src/components/list_items_panel/list_tree_item.tsx
```diff
@@ -426,14 +409,18 @@ export class DatasetResource extends Resource<DatasetProps> {
...
       this.setState({ loading: true });
-      await listTablesService.listTables(dataset.id).then((data: Dataset) => {
-        newDataset.tables = data.tables;
-        newDataset.tableIds = data.tableIds;
-      });
-      await listModelsService.listModels(dataset.id).then((data: Dataset) => {
-        newDataset.models = data.models;
-        newDataset.modelIds = data.modelIds;
-      });
+      await bigQueryService
```
prodonjs, line 412: Since you're using `await` here, the listTables and listModels API calls will run sequentially when it looks like you can probably run them concurrently and improve the overall performance. 

There's sort of an odd blend of newer async/await style Promise handling along with older `.then(` based handling. The latter is actually probably bit more intuitive in this case where you want each Promise to run its asynchronous code concurrently and then update the state when it finishes.

At a high level, I think you want to issue both requests, handle errors in either one without affecting the other, and when both have completed (either successfully or with error) set the loading state to false.

jupyterlab_bigquery/src/components/list_items_panel/service/bigquery_service.ts
```diff
@@ -63,4 +45,77 @@ export class BigQueryService {
         };
       });
   }
+
+  async listTables(projectId, datasetId): Promise<Partial<Dataset>> {
```
prodonjs, line 49: You have the async modifier on the method but you use the older `.then(` based Promise handling. In general, it's best not to mix the two. I think async/await is more readable since you can use the more traditional try/catch block to handle any errors and you don't have to worry about callbacks after your asynchronous method resolves its data.

jupyterlab_bigquery/src/components/list_items_panel/service/bigquery_service.ts
```diff
@@ -63,4 +45,77 @@ export class BigQueryService {
...
+          };
+        }
+        return {
+          tables: tables,
+          tableIds: tableIds,
+        };
+      });
+  }
+
+  async listModels(projectId, datasetId): Promise<Partial<Dataset>> {
```
prodonjs, line 89: Same as above.

prodonjs (changes requested):

jupyterlab_bigquery/src/components/list_items_panel/service/bigquery_service.ts
cbwilkes: Removed the `then` based Promise handling

jupyterlab_bigquery/src/components/list_items_panel/list_tree_item.tsx
```diff
@@ -409,18 +409,24 @@ export class DatasetResource extends Resource<DatasetProps> {
...
+
+      const modelsResult =
+        bigQueryService
         .listModels(dataset.projectId, dataset.name)
         .then((data: Dataset) => {
           newDataset.models = data.models;
           newDataset.modelIds = data.modelIds;
         });
+
+      await tablesResult;
```
prodonjs, line 428: You could also do `await Promise.all([tablesResult, modelsResult])`. Regardless of whether you use the existing code or that, you may want to add `.catch(` handlers to each Promise to not throw an exception if the list request fails.

prodonjs (approved):
Looks good. One suggestion if we want to be able to handle one of the requests failing but allowing the other to succeed, but if we're OK with all or nothing, we can proceed as is.

prodonjs (approved):

cbwilkes:
Unfortunately, allowing one request to succeed would change the behavior. We would have to change the interface to let the user know that models or tables didn't succeed. Currently, the user wouldn't know that data assets are missing from the list.

PR #404, opened 2021-01-21
More UX fine tuning

mounikasomisetty:
https://screencast.googleplex.com/cast/NjY0NzcwODQyNzIyMzA0MHxhZGRkMGNhNC03Zg
+ Fixed spacing in side bar header and view result button
+ dividers full length fix
+ week scheduler spacing fix
+ bump shared version

### DIFF
--- a/jupyterlab_gcpscheduler/package.json
+++ b/jupyterlab_gcpscheduler/package.json
@@ -52,7 +52,7 @@
     "@phosphor/signaling": "^1.3.0",
     "csstips": "^1.2.0",
     "formik": "^1.5.8",
-    "gcp_jupyterlab_shared": ">=1.1.2",
+    "gcp_jupyterlab_shared": "^1.1.2",
     "typestyle": "^2.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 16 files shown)


jupyterlab_gcpscheduler/package.json
```diff
@@ -52,7 +52,7 @@
     "@phosphor/signaling": "^1.3.0",
     "csstips": "^1.2.0",
     "formik": "^1.5.8",
-    "gcp_jupyterlab_shared": "1.1.2",
+    "gcp_jupyterlab_shared": ">=1.1.2",
```
prodonjs, line 55: I think using caret `^1.1.2` is best here since it will ensure we remain on the same major version.

prodonjs (changes requested):
Just a small suggestion in the `package.json` file. Also, let's switch the Badge to use _Experimental_ instead of Alpha since that's the nomenclature we are using these days.

jupyterlab_gcpscheduler/package.json
mounikasomisetty: Ah okay. I thought since 1.1.2 had a breaking change (more icons) it had to be >=. Changed!

mounikasomisetty:
http://screen/3rcEEKDtQ5mxocN
http://screen/ANXc9yXJbrSMkKE
Changed 'alpha' to experimental

prodonjs (changes requested):
LGTM but I think you may need to update some snapshots after the badge change.

prodonjs (approved):

PR #406, opened 2021-01-22
Order executions and schedules by name

mounikasomisetty:
Adding orderBy 'name desc' to listExecutions and listSchedules so at least the order is consistent (& tests)


prodonjs:
Superseded by #407

PR #407, opened 2021-01-25
Order executions and schedules by name

prodonjs:
This is a copy of #406 that fixes the failing test.


prodonjs:
@angelakuo - 

Creating this PR from #406 since @mounikasomisetty is out for the first half of the week and we'd like to proceed with getting all necessary Executor changes merged. This just fixes the failing test from that PR.

PR #432, opened 2021-06-29
Remove unsupported extension packages

prodonjs:
All of these extensions are now out-of-date and unsupported as we've moved the code elsewhere. Removing them to avoid any ongoing confusion or support questions.


angelakuo (approved):

em-pe:
HI @prodonjs, could you shed some light on where were those extensions moved? I was trying to find them on github, but had no luck with that.

PR #492, opened 2024-02-26
kernels-mixer: Fix a bug where the remote kernel doesn't refresh the state

ywskycn:


ojarjur:
Can we reproduce the failure in the [`TestKernelModel` test](https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/630f4a69b2b646638b51003d7817f5f3afa9b6f2/kernels-mixer/kernels_mixer/kernels_test.py#L132) to confirm that this fixes it?

That might require some fixes to the mocked classes in that same file, but they're extending the real code from the  `jupyter_server` repo that is used in practice, so I think it should be doable.

ywskycn (commented):
fix test

ojarjur (dismissed):

ojarjur (approved):

PR #497, opened 2024-05-02
[kernels-mixer] Don't list remote kernels unless the user has created one

ojarjur:


kernels-mixer/kernels_mixer/kernels_test.py
```diff
@@ -115,12 +116,15 @@ def _get_changed_ports(self, *args, **kwargs):
...
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, list_delay=None, **kwargs):
         self.kernel_spec_manager = MockKernelSpecManager()
         self.kernel_spec_manager.default_kernel_name = remote_kernel
         self.kernel_manager_class = "kernels_mixer.kernels_test.MockKernelManager"
+        self._list_delay = list_delay
 
     async def list_kernels(self, **kwargs):
+        if self._list_delay:
+            time.sleep(self._list_delay)
```
ptwng, line 127: should this be something like await asyncio.sleep?  Otherwise should this still be an async function?

ptwng (approved):

PR #524, opened 2026-04-08
Generate auth tokens ahead of time to avoid blocking the event loop

ajma:
The Jupyter server uses a single-threaded Tornado event loop, and the gateway client calls the token renewer's `get_token()` on that thread whenever it needs to authenticate with the kernel gateway.

Every `token_lifetime_seconds` (300 by default), `get_token()` called `force_new_token()`, which ran `subprocess.run("gcloud ...")` and waited for it — 1-5 seconds during which the event loop is frozen. No HTTP requests served, no WebSocket messages processed, no notebooks responding. The server appears hung.

`get_token()` is synchronous by contract (`GatewayTokenRenewerBase`), so it can't just `await` an async subprocess. But it doesn't need to: `subprocess.Popen` already runs the child concurrently, and only `.communicate()` blocks. So run the command ahead of time and collect the result later.

`CommandTokenRenewer` now holds a `Popen` for an already-running token command:

1. `force_new_token()` calls `.communicate()` on it — this returns immediately, because the command was started at the previous renewal, 300 seconds ago.
2. It checks the return code and raises `CalledProcessError` on failure (replacing `check=True`).
3. It starts the process that will supply the *next* token.

The constructor starts the first process, so the initial token is generated in parallel with the rest of the server's startup rather than blocking it.

`CachedTokenRenewerBase` is unchanged — no threads, no locks, no shared mutable state.

One tradeoff worth naming: a token is now generated up to one renewal interval before it is used, so its maximum age at use goes from 300s to ~600s. The default token command requests `--min-expiry=30m`, so that is still a wide margin.

Also adds `tokenrenewer_test.py`; the module had no tests. The three tests covering ahead-of-time generation fail against the current `master` implementation.


ojarjur:
I think the insight is valid, but the proposed fix is more complicated than it needs to be because the wrong class is being modified.

We don't need to spawn a new thread to run the subprocess invocations in because subprocess calls already run asynchronously.

Instead, all we have to do is make the `CommandTokenRenewer` class not block on the result of the subprocess invocation.

I.E. create a `subprocess.Popen` object instead of calling `subprocess.run`.

We can have it store a field for the previously created `Popen` object, and then in the body of the `force_new_token` method make it do the following:

1. Call `.communicate()` on the previously created `Popen` object to get its result and stdout.
2. Check the result to make sure the command succeeded.
3. Update the stored field with a newly constructed `Popen` object that runs the configured command with `stdout=PIPE`.

Addditionally, the constructor of the `CommandTokenRenewer` class can initialize the stored field with the initial `Popen` object, so that even the first call to `get_token` doesn't have to block for the full time the subprocess is running.

ajma:
Reworked to use `subprocess.Popen` in `CommandTokenRenewer` as suggested — `CachedTokenRenewerBase` is now untouched.

`force_new_token()` calls `.communicate()` on the `Popen` started during the previous renewal (so it returns immediately), checks the return code, then starts the process for the next renewal. The constructor starts the first one, so startup isn't blocked either. This also let the `tempfile` go away: with `stdout=PIPE` and `encoding="UTF-8"`, `communicate()` hands back a `str` directly.

Two things the thread-based version got wrong that this avoids:

- **It returned the placeholder token on the first call.** `configure_gateway_client` sets `GatewayClient.auth_token` to `"Initial, invalid value"` (the workaround for jupyter-server/jupyter_server#1339), and `gateway_client.py` passes that straight into `get_token()`. Because the base class treated any non-empty `auth_token` as "stale but still valid", the first call returned the placeholder and merely *scheduled* a real fetch.
- **Failures were silent.** The background thread caught and discarded the `CalledProcessError`, so a broken token command would just keep serving a stale token indefinitely. The explicit return-code check preserves the old `check=True` behavior.

Also added `tokenrenewer_test.py`, which the module didn't have. The three tests covering the ahead-of-time behavior fail against the current `master` implementation, and `test_placeholder_auth_token_is_replaced` fails against the thread-based version. Green on all 12 matrix configs.

PR #525, opened 2026-04-08
Reuse a single ProcessPoolExecutor instead of creating one per call

ajma:
On Windows, a new ProcessPoolExecutor was created and destroyed on every gcloud invocation, adding ~100-500ms of process spawn overhead each time. Use a module-level singleton instead.

### DIFF
--- a/google-cloud-jupyter-config/google/cloud/jupyter_config/config.py
+++ b/google-cloud-jupyter-config/google/cloud/jupyter_config/config.py
@@ -54,10 +54,14 @@ def run_gcloud_subcommand(subcmd):
         return t.read().decode("UTF-8").strip()
 
 
+_process_pool_executor = futures.ProcessPoolExecutor(max_workers=1)
+
+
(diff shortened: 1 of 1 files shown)


ojarjur:
This looks like a good catch to me; thanks!

I was wondering if we can reproduce/confirm the performance gains with the presubmit tests, so I compared the timing of the test runs against the 4 most recent closed PRs.

All of the previous runs (for Windows with the latest Python version) completed in between 21 and 25 seconds, whereas this one completed in 16 seconds.

I'm going to manually retry the presubmits a few times and if the timing consistently remains faster, then I think we'll have solid evidence this change achieves the desired performance improvements.

ojarjur:
The subsequent runs completed in:

* 15 seconds
* 15 seconds
* 14 seconds

... so, it seems pretty clear that the performance improvement is consistent and quite large.

ojarjur (approved):

PR #534, opened 2026-08-17
[google-cloud-jupyter-config] Fix bug in calling CalledProcessError

ojarjur:
Previously, were were passing too many arguments to the `CalledProcessError` constructor.

I had seen from [the docs](https://docs.python.org/3/library/subprocess.html#subprocess.CalledProcessError) that it has 5 fields, and mistakenly thought that meant it takes 5 arguments in the constructor, but in actuality the `stdout` field is just an alias for `output`, and it really only takes 4 arguments (returncode, cmd, output, and stderr).


dparikh:
LGTM

tim-u (approved):

PR #536, opened 2026-08-18
Add support for notification propagation

tim-u:
# [google-cloud-jupyter-config] Add notification system support for Dataproc kernels

Introduces the backend infrastructure to capture and emit Dataproc kernel warnings and errors as Jupyter Server events (`event_logger`). This allows frontend extensions (such as `dataproc-jupyter-plugin`) to surface real-time actionable notifications (e.g., deprecated regions, dead kernels, quota/permission failures, and WebSocket connection errors) to the user.

  * Added DataprocGatewayKernelSpecManager that intercepts `warnings` in `list_kernel_specs()` and forwards them to the notification sink.
  * Added DataprocGatewayMappingKernelManager that intercepts dead/unresponsive kernels in `list_kernels()` and forwards failure reasons to the notification sink.
  * Added DataprocGatewayWebSocketConnection that catches connection exceptions on remote WebSocket channels and reports error details to the active sink. Includes fallback inheritance to `GatewayWebSocketConnection` when kernels-mixer is not present.
  * Implemented DataprocNotificationHandler which ignores duplicate notifications and emits the events .
  * Defined and registered the notification event schema, wired `DataprocNotificationHandler` into server kernel managers.


ajma:
### Code review

Found 2 issues:

1. Kernel-spec and dead-kernel notifications never fire under kernels-mixer, because `configure_gateway_client` only wires the WebSocket path and misses the equivalent config for the other two managers.

`config.py` sets `c.DelegatingWebsocketConnection.remote_websocket_connection_class` to the new Dataproc websocket class, but never sets `c.MixingKernelSpecManager.remote_kernel_spec_manager_class` or `c.MixingMappingKernelManager.remote_kernel_manager_class` (both configurable traits, added in #537 and already on `master`). Under a real kernels-mixer deployment, `server_app.kernel_spec_manager.remote_manager` / `server_app.kernel_manager.remote_manager` stay as the default plain `GatewayKernelSpecManager` / `GatewayMappingKernelManager`, which have no `notifications_sink` attribute — so the `hasattr(spec_mgr, "notifications_sink")` checks in `__init__.py` silently no-op and `DataprocGatewayKernelSpecManager`/`DataprocGatewayMappingKernelManager` are never instantiated. Only the WebSocket-failure notification path actually works in that configuration. `notifications_test.py::test_extension_initialization_with_mixer` masks this because it mocks `remote_manager` with `MagicMock(spec=["notifications_sink"])` instead of a real mixer manager.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/936eedd27c8a1ba9a602619bd808b9b325496d2e/google-cloud-jupyter-config/google/cloud/jupyter_config/config.py#L254-L262

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/936eedd27c8a1ba9a602619bd808b9b325496d2e/google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py#L59-L70

2. `DataprocNotificationHandler.seen_ids` is a plain `set`, so the "purge oldest half" eviction doesn't purge the oldest anything.

`self.seen_ids` has no defined iteration order, so `set(list(self.seen_ids)[to_remove:])` discards an arbitrary, hash-order-dependent half rather than the oldest entries. Once the 1000-entry cap is first hit, a notification ID added moments ago can be evicted while one seen long ago is kept — so a previously-emitted (and possibly already-dismissed) notification can resurface unpredictably on a later poll, defeating the dedup this class exists for.

https://github.com/GoogleCloudPlatform/jupyter-extensions/blob/936eedd27c8a1ba9a602619bd808b9b325496d2e/google-cloud-jupyter-config/google/cloud/jupyter_config/notifications.py#L69-L76

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tim-u:
> 1. Kernel-spec and dead-kernel notifications never fire under kernels-mixer, because `configure_gateway_client` only wires the WebSocket path and misses the equivalent config for the other two managers.

That's WAI, `google-cloud-jupyter-config` is not meant to have a dependency on the kernels-mixer, so it shouldn't be aware of either MixingKernelSpecManager or MixingMappingKernelManager. In `dataproc-jupyter-plugin` we [will set both Mixing* traits](https://github.com/GoogleCloudDataproc/dataproc-jupyter-plugin/pull/444), since that client uses both `kernels-mixer` and `google-cloud-jupyter-config`.

Update: Just in case we need to use the `google-cloud-jupyter-config` without the `dataproc-jupyter-plugin`, I'm also setting the managers in `google-cloud-jupyter-config`.

> 2. DataprocNotificationHandler.seen_ids is a plain set, so the "purge oldest half" eviction doesn't purge the oldest anything.

Fixed.

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
```diff
@@ -20,6 +20,29 @@
...
+from google.cloud.jupyter_config.managers import DataprocGatewayKernelSpecManager
+from google.cloud.jupyter_config.managers import DataprocGatewayMappingKernelManager
+from google.cloud.jupyter_config.websockets import DataprocGatewayWebSocketConnection
+from google.cloud.jupyter_config.notifications import (
+    NOTIFICATION_SCHEMA_ID,
+    DATAPROC_NOTIFICATION_SCHEMA,
+    DataprocNotificationHandler,
+)
+
+__all__ = [
```
ojarjur: What's the motivation behind this?

We normally forbid imports of the form `from ... import *`, so I don't believe this would have any direct impact on any of our code.

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
```diff
@@ -28,6 +51,28 @@ def _load_jupyter_server_extension(server_app):
...
     server_app.web_app.add_handlers(host_pattern, [(config_url, PropertiesHandler)])
 
+    try:
+        server_app.event_logger.register_event_schema(DATAPROC_NOTIFICATION_SCHEMA)
+
+        handler = DataprocNotificationHandler(server_app.event_logger)
+
+        # Wire the sink directly, unpacking mixing managers if present
+        spec_mgr = server_app.kernel_spec_manager
+        if hasattr(spec_mgr, "remote_manager"):
```
ojarjur: I think this is sensitive to the order in which different extensions are loaded.

For example, if this extension is loaded before the `dataproc_jupyter_plugin` extension is loaded, then the `kernel_spec_manager` will not yet have been updated, so this logic will not kick in.

The way this is currently setup requires two steps:

1. An extension (or user config) sets the `kernel_spec_manager` appropriately
2. This extension then sets the `notifications_sink` on that kernel_spec_manager.

That, in turn, requires everything to be loaded in only a specific order, which we cannot ensure.

I think it would be better to instead have the `DataprocNotificationHandler` be a subclass of [`SingletonConfigurable`](https://traitlets.readthedocs.io/en/stable/config-api.html#traitlets.config.SingletonConfigurable) and then have the various managers call the `DataprocNotificationHandler.instance()` method to get the instance of the sink.

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
tim-u: Removed

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
tim-u: I did some manual testing and it seemed to work fine, but to avoid the fragility I've converted the `DataprocNotificationHandler` to a singleton.

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
```diff
@@ -28,6 +37,17 @@ def _load_jupyter_server_extension(server_app):
     config_url = url_path_join(base_url, "gcloud", "config", "properties")
     server_app.web_app.add_handlers(host_pattern, [(config_url, PropertiesHandler)])
 
+    try:
+        server_app.event_logger.register_event_schema(DATAPROC_NOTIFICATION_SCHEMA)
+        parent = server_app if isinstance(server_app, Configurable) else None
+        DataprocNotificationHandler.instance(
+            parent=parent,
+            event_logger=getattr(server_app, "event_logger", None),
```
ojarjur, line 45: We just referenced the `event_logger` attribute of the `server_app` value up on line 41.

That means this `getattr` call is unnecessary here; either the `event_logger` attribute is defined and can be directly referenced, or the reference up on line 41 would have already raised an error.

I would say to either get rid of this call to `getattr`, or move it up to before line 41 and have the call to `register_event_schema` only happen if the attribute was found.

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
```diff
@@ -0,0 +1,119 @@
...
+      ),
+  )
+
+  def _get_notifications_sink(self):
+    if self.notifications_sink is not None:
+      return self.notifications_sink
+    if DataprocNotificationHandler.initialized():
+      return DataprocNotificationHandler.instance()
+    return None
+
```
ojarjur, line 44: From `traitlets` import `default` and then this can be simplified to:

```py
    @default("notifications_sink")
    def _default_notifications_sink(self):
        return DataprocNotificationHandler.instance()
```

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
```diff
@@ -0,0 +1,119 @@
...
+    GatewayKernelSpecManager,
+    GatewayMappingKernelManager,
+)
+from traitlets import Any
+
+
+class DataprocGatewayKernelSpecManager(GatewayKernelSpecManager):
+  """A GatewayKernelSpecManager that intercepts kernelspec notifications."""
+
+  notifications_sink = Any(
```
ojarjur, line 29: This should be a `traitlets.Callable` instead of `traitlets.Any`: https://traitlets.readthedocs.io/en/stable/trait_types.html#traitlets.Callable

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
```diff
@@ -0,0 +1,119 @@
...
+
+    if notifications and sink:
+      sink(notifications)
+    return kernel_specs
+
+
+class DataprocGatewayMappingKernelManager(GatewayMappingKernelManager):
+  """A GatewayMappingKernelManager that intercepts kernel notifications."""
+
+  notifications_sink = Any(
```
ojarjur, line 78: Same comment as above about this being a `Callable`

google-cloud-jupyter-config/google/cloud/jupyter_config/__init__.py
tim-u: Done

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
tim-u: Done

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
tim-u: Done

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
tim-u: Done

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
```diff
@@ -0,0 +1,113 @@
...
+
+    sink = self.notifications_sink
+    if not isinstance(kernels, list) or not sink:
+      return kernels
+
+    for kernel in kernels:
+      if isinstance(kernel, dict) and (
+          kernel.get("execution_state") == "dead" or kernel.get("status") == "dead"
+      ):
+        kernel_id = kernel.get("id", "unknown")
```
ojarjur: If for some reason we got kernels with no `id` field, then this would combine the failures for all of them together.

Instead, let's do something like taking a hash of the full kernel contents so that distinct, dead (but with no-id) kernels are better distinguished.

... something like:

```py
kernel_id = kernel.get("id", None) or hashlib.sha256(json.dumps(kernel)).hexdigest()
```

google-cloud-jupyter-config/google/cloud/jupyter_config/websockets.py
```diff
@@ -0,0 +1,64 @@
...
+  def _report_websocket_event(self, message):
+    """Find the notifications_sink in the manager hierarchy and report WebSocket event."""
+    if not DataprocNotificationHandler.initialized():
+      return
+
+    sink = DataprocNotificationHandler.instance()
+    try:
+      sink([
+          {
+              "id": f"ws-{self.kernel_id}-{abs(hash(message))}",
```
ojarjur: `abs(hash(message))` is a lot more likely to encounter hash conflicts than if we used hashlib:

```{hashlib.sha256(message).hexdigest()}```

tim-u: Done

google-cloud-jupyter-config/google/cloud/jupyter_config/managers.py
tim-u: Done

ojarjur (approved):

PR #538, opened 2026-09-10
[kernels-mixer] better support for reconnections to kernels.

ojarjur:
This commit adds four improvements for when the client reconnects to a kernel:

1. Don't send the `starting` status message for kernels that have already communicated with the client.

   This fixes an issue where the kernel status in the UI can change from `idle` or `busy` to `starting`.

2. Send the `kernel_info_request` used to health check the connection on the `control` channel instead of the `shell` channel.

   This prevents an issue where a large amount of shell messages coming back from the kernel (such as large amounts of stderr) can flood the `shell` channel and as a result block the `kernel_info_reply` message from coming back in a reasonable time.

3. Buffer any non `kernel_info_reply` messages seen during the connection health check and play them back after the connection is returned to the caller.

   This reduces the risk of messages getting dropped during the period between when a connection is dropped and the next one is created.

4. Poll for the `kernel_info_reply` for a fixed amount of time rather than a fixed number of messages.

    This similarly makes us more gracefully handle the scenario where the
    channel is being flooded with a very large number of messages.

### DIFF
--- a/kernels-mixer/kernels_mixer/websockets.py
+++ b/kernels-mixer/kernels_mixer/websockets.py
@@ -15,6 +15,9 @@
 import asyncio
 import datetime
 import json
+import queue
+import threading
+import time
 import uuid
 
 import tornado.gen as tornado_gen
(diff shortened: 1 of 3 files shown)


kernels-mixer/kernels_mixer/websockets.py
```diff
@@ -29,6 +32,36 @@
...
+class _BufferedConnection(tornado_websocket.WebSocketClientConnection):
+    """Object that wraps a WebSocketClientConnection and adds a buffer of previous messages to replay.
+    """
+    def __init__(self, conn: tornado_websocket.WebSocketClientConnection, buffered_messages: queue.Queue):
+        self._conn = conn
+        self._buffered_messages = buffered_messages
+
+    @property
+    def selected_subprotocol(self):
+        return conn.selected_subprotocol
```
tim-u, line 44: `self._conn`?

kernels-mixer/kernels_mixer/websockets.py
```diff
@@ -29,6 +32,36 @@
...
+        return self._conn.write_message(*args, **kwargs)
+
+    async def read_message(self, callback=None):
+        if not self._buffered_messages.empty():
+            msg = self._buffered_messages.get()
+            if callback:
+                callback(msg)
+                return None
+            return msg
+        return await self._conn.read_message()
```
tim-u, line 62: Do we need to pass the callback to the original `read_message()` if `_buffered_messages` is empty?

kernels-mixer/kernels_mixer/websockets.py
```diff
@@ -164,30 +208,45 @@ def __init__(self, *args, **kwargs):
             jupyter_server.gateway.connections.tornado_websocket = _InterceptedTornadoWebsocket(self.log)
             StartingReportingWebsocketConnection.patched_websocket_connect = True
 
+    def has_seen_kernel_activity(self):
+        with StartingReportingWebsocketConnection.seen_kernels_lock:
+            return self.kernel_id in StartingReportingWebsocketConnection.seen_kernels
+
+    def record_kernel_activity(self):
```
tim-u, line 215: From Gemini:
> Because handle_outgoing_message unconditionally calls self.record_kernel_activity(), this block executes for every single message sent over the websocket. For kernels that generate a lot of rapid output (e.g., printing in a loop or progress bars), iterating over a list of up to 1000 items and creating a new list on every message will cause significant CPU overhead and garbage collection churn.

kernels-mixer/kernels_mixer/websockets.py
ojarjur: Done

kernels-mixer/kernels_mixer/websockets.py
ojarjur: Changed to use a dictionary instead of a list and to only do the cleanup when we actually hit the size limit.

The size limit is large enough that most users will never hit it, but even when they do hit it it will be a very rare occasion.

tim-u (approved):

